PR #196, opened 2023-07-19
IOS-1499 Event Converter | Reorganise some converters

mgolovko:


sachnik (approved):

Modules/Services/Sources/Models/Block/BlockContent/Content/BlockRelation.swift
```diff
@@ -1,7 +1,7 @@
 import Foundation
 
 public struct BlockRelation: Hashable {
-    public let key: String
```
bududomasidet, line 4: It might be alright to change these variable to a mutating ones due to legacy logic for handling all middleware updates.
Imho for better handling errors and converter issues we need to leave it immutable.

bududomasidet (approved):

Modules/Services/Sources/Models/Block/BlockContent/Content/BlockRelation.swift
mgolovko: What advantage do you see in updating via init? We have a public init, we don't blocking the rewrite, but make rewrite more harder.

Anytype/Sources/Models/Documents/Events/EventConverters/MiddlewareEventConverter.swift
```diff
@@ -138,126 +138,26 @@ final class MiddlewareEventConverter {
...
             return .general
             
-        case let .blockSetFile(newData):
-            guard newData.hasState else {
+        case let .blockSetFile(data):
+            infoContainer.updateFile(blockId: data.id) { $0.handleSetFile(data: data) }
+            
+            if data.hasState {
                 return .general
             }
```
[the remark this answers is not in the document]
mgolovko: oh, yes, i should using `guard`

Modules/Services/Sources/Models/Block/BlockContent/Content/BlockRelation.swift
bududomasidet: you can be confident about your model hasn't been updated somewhere in code. And it could be created only in one place -> where we receiving value from middleware

PR #198, opened 2023-07-19
iOS-1523 Trim typed/inserted text in recovery phrase text view

sachnik:
Validate typing and inserting recovery phrase - max 12 words + 8 characters per word, place it under the toggle.
Not so obvious in my opinion but we want to try how it will work.

https://github.com/anyproto/anytype-swift/assets/6065952/0b678fc5-7409-47fd-b1f3-809108e9449c


bududomasidet (approved):

mgolovko:
I agree. How user should be understand logic, 8 characters per word.... We can make red or make something, that user can understand.

mgolovko (approved):

sachnik:
> I agree. How user should be understand logic, 8 characters per word.... We can make red or make something, that user can understand.
I think we will improve this logic later, agree with you

PR #202, opened 2023-07-20
iOS-1304 Create ObjectSubscriptionService and use it in the SubscriptionToggler

sachnik:


Anytype/Sources/PresentationLayer/TextEditor/Set/Views/Collection/SetCollectionView.swift
```diff
@@ -1,4 +1,5 @@
 import SwiftUI
+import Services
```
bududomasidet, line 2: is it possible to not disclose services in view layer?

Anytype/Sources/PresentationLayer/TextEditor/Set/Views/Table/SetTableView.swift
```diff
@@ -1,4 +1,5 @@
 import SwiftUI
+import Services
```
bududomasidet, line 2: and here

Modules/Services/Sources/Models/SubscriptionData/SubscriptionData.swift
```diff
@@ -1,47 +1,36 @@
...
-        }
-    }
-    
-    var rowsPerPage: Int {
-        switch self {
-        case .search(let data):
-            return data.limit
-        case .objects:
-            return UserDefaultsConfig.rowsPerPageInSet
+    public static func == (lhs: SubscriptionData, rhs: SubscriptionData) -> Bool {
```
bududomasidet, line 5: it seems to me this function is redundant

Modules/Services/Sources/Models/SubscriptionData/SubscriptionId.swift
```diff
@@ -0,0 +1,7 @@
+public struct SubscriptionId: Hashable {
```
bududomasidet, line 1: Do we really need this structure? Why not just use string or typealias?

mgolovko: I think we can delete it

mgolovko (dismissed):

Anytype/Sources/PresentationLayer/TextEditor/Set/Views/Collection/SetCollectionView.swift
joe-pusya: Import was removed due to SubscriptionId deletion here and below

Modules/Services/Sources/Models/SubscriptionData/SubscriptionId.swift
joe-pusya: SubscriptionId was deleted

mgolovko (approved):

PR #209, opened 2023-07-24
IOS-1587 Template selection. Change alert type.

bududomasidet:
Screenshot 2023-07-24 at 18 30 17


Anytype/Sources/PresentationLayer/Templates/Views/Selection/Objects/TemplatePreview.swift
```diff
@@ -63,7 +63,7 @@ struct TemplatePreview: View {
     
     private var cover: some View {
         Group {
-            if case let .installed(templateModel) = viewModel.model {
+            if case let .installed(templateModel) = viewModel.model.model {
```
mgolovko, line 66: Maybe rename internal `model.model` for better understanding?

PR #222, opened 2023-07-26
IOS-1579 Editor performance temporary hotfix

bududomasidet:
@mgolovko I want to merge it to the next release because it is super inconvenient to type something in editor. 
While we are still in researching state of how to refactor eventBunch & whole EditorPageController & blocks usage I would like to make it possible to write long texts in editor


Anytype/Sources/ServiceLayer/Block/Text/TextService.swift
```diff
@@ -10,7 +10,12 @@ final class TextService: TextServiceProtocol {
             $0.blockID = blockId
             $0.text = middlewareString.text
             $0.marks = middlewareString.marks
-        }).invoke()
+        }).invoke(shouldHandleEvent: false)
```
mgolovko, line 13: block text set receive a empty events. For what this?

PR #234, opened 2023-08-02
IOS-1518 Add layout check before template creation

bududomasidet:


joe-pusya (approved):

Modules/Services/Sources/Services/Templates/File.swift
```diff
@@ -17,11 +17,22 @@ public final class TemplatesService: TemplatesServiceProtocol {
     }
     
     public func createTemplateFromObjectType(objectTypeId: BlockId) async throws -> BlockId {
```
mgolovko, line 19: Swift file name is 'File.swift'

mgolovko (approved):

Modules/Services/Sources/Services/Templates/File.swift
bududomasidet: to be resolved later

PR #235, opened 2023-08-02
iOS-1202 Colourful recovery phrase

joe-pusya:
Unfortunately I didn’t have time to do it with TextKit, so it’s a simple version for now, hope to remake it in the future because this version has some hacks and limitations

https://github.com/anyproto/anytype-swift/assets/6065952/a2b7198a-5f81-4c1c-9aec-671ca00a9b89


mgolovko (approved):
beautifully 🟢🟡🟣🔴🔵

PR #240, opened 2023-08-03
IOS-1516 Tiny improvements

bududomasidet:


Anytype/Sources/PresentationLayer/TextEditor/EditorPage/Views/Header/Filled/Subviews/ObjectHeaderView.swift
```diff
@@ -204,7 +203,7 @@ private extension ObjectHeaderView {
                 excluding: [.bottom],
                 insets: .zero
             )
-            coverBottomConstraint = $0.bottom.greaterThanOrEqual(
```
bududomasidet, line 207: it causes to warning "Unable to simultaneously satisfy constraints" but I don't have an idea why greaterThanOrEqual doesn't work with SwiftUI properly

PR #263, opened 2023-08-10
iOS-1649 Fix collection header availability

joe-pusya:


bududomasidet (approved):
thank you @joe-pusya ! I haven't seen this problem :)

PR #304, opened 2023-08-31
IOS-1705 Core | add thread-safe array in storages

mgolovko:


Modules/AnytypeCore/AnytypeCore/Utils/Synchronized colections/SynchronizedDictionary.swift
```diff
@@ -45,4 +45,10 @@ public final class SynchronizedDictionary<K, V> where K: Hashable {
         
         return keys
     }
+    
+    public func removeAll() {
```
joe-pusya, line 49: just of the future?

joe-pusya (approved):

Modules/AnytypeCore/AnytypeCore/Utils/Synchronized colections/SynchronizedDictionary.swift
mgolovko: I using it in ObjectTypeProvider

joe-pusya: ok)

PR #313, opened 2023-09-04
iOS-1681 | Sharing extension

bududomasidet:


Anytype/Sources/Design system/InlinePicker/AnytypeInlinePicker.swift
```diff
@@ -0,0 +1,49 @@
...
+    @State private var initialValue: T?
+    private var allValues: [T]
+    private let selectionHandler: (T) -> Void
+    
+    public init(allValues: [T], selectionHandler: @escaping (T) -> Void) {
+        self.allValues = allValues
+        self.selectionHandler = selectionHandler
+    }
+    
+    @MainActor public var body: some View {
```
mgolovko, line 17: body is mainActor by default

Anytype/Sources/PresentationLayer/ShareView/ShareModuleAssembly.swift
```diff
@@ -0,0 +1,61 @@
...
+            return nil
+        }
+        
+        let viewModel = ShareViewModel(
+            contentViewModel: contentViewModel,
+            interactor: serviceLocator.sharedContentInteractor,
+            contentManager: serviceLocator.sharedContentManager
+        )
+    
+        viewModel.onClose = onClose
```
mgolovko, line 57: Why do you setup part from init and part from properties?

Anytype/Sources/PresentationLayer/ShareView/ShareViewModelInteractor.swift
```diff
@@ -0,0 +1,161 @@
...
+                targetId: lastBlockInDocument,
+                info: blockInformation,
+                position: .bottom
+            )
+        }
+    }
+}
+
+
+extension URL {
```
mgolovko, line 130: maybe mark this extensions as private?

Anytype/Sources/PresentationLayer/ShareView/SharedContentManager.swift
```diff
@@ -0,0 +1,41 @@
...
+    case image(URL)
+}
+
+final class SharedContentManager: SharedContentManagerProtocol {
+    private lazy var encoder = JSONEncoder()
+    private lazy var decoder = JSONDecoder()
+    private lazy var userDefaults = UserDefaults(suiteName: TargetsConstants.appGroup)
+    
+    func saveSharedContent(content: [SharedContent]) throws {
+        let sharedData = try JSONEncoder().encode(content)
```
mgolovko, line 21: `try encoder.encode(content)`?

Anytype/Sources/ServiceLayer/Object/Search/SearchService.swift
```diff
@@ -3,8 +3,11 @@ import Combine
 import Services
 import AnytypeCore
 
-protocol SearchServiceProtocol: AnyObject {
+protocol SearchInteractorProtocol: AnyObject {
```
mgolovko, line 6: why Interactor? this is service 🤔

Anytype/Sources/PresentationLayer/ShareView/ShareModuleAssembly.swift
bududomasidet: I believe it's essential to encapsulate routing logic separately from assembly concerns. There are instances when we may need to invoke a method on the viewModel after a routing operation, and I don't believe assembly should be responsible for handling such tasks.

PR #339, opened 2023-09-08
iOS-1641 Make modules for the views list in Set / Collection and update design

joe-pusya:


Anytype/Sources/PresentationLayer/TextEditor/Set/Views/Popups/ViewPicker/SetViewPicker.swift
```diff
@@ -1,12 +1,20 @@
 import SwiftUI
 
-struct EditorSetViewPicker: View {
-    @ObservedObject var viewModel: EditorSetViewPickerViewModel
+struct SetViewPicker: View {
+    @StateObject var viewModel: SetViewPickerViewModel
```
bududomasidet, line 4: Screenshot 2023-09-11 at 13 23 41

bududomasidet (approved):

PR #361, opened 2023-09-18
iOS-1733 Core | Remove toggles

mgolovko:


Anytype/Sources/PresentationLayer/TextEditor/Set/EditorSetViewModel.swift
```diff
@@ -177,10 +177,6 @@ final class EditorSetViewModel: ObservableObject {
         setDocument.setUpdatePublisher.sink { [weak self] in
             self?.onDataChange($0)
         }.store(in: &subscriptions)
-
-        Publishers.CombineLatest(setDocument.document.detailsPublisher, $isAppear)
-            .sink { [weak self] in self?.handleDetails(details: $0, isAppear: $1) }
-            .store(in: &subscriptions)
```
joe-pusya, line 183: do we need to maintain `isAppear` for any other reason?

joe-pusya:
👍

Anytype/Sources/PresentationLayer/TextEditor/Set/EditorSetViewModel.swift
mgolovko: No. Deleted it

PR #373, opened 2023-09-25
IOS-1314 Common | Delete migration guide toggle

mgolovko:


Anytype/Sources/PresentationLayer/Auth/LoginFlow/LoginViewModel.swift
```diff
@@ -183,11 +183,7 @@ final class LoginViewModel: ObservableObject {
                     errorText = Loc.accountDeleted
                 }
             } catch SelectAccountError.failedToFindAccountInfo {
-                if FeatureFlags.migrationGuide {
-                    output?.onShowMigrationGuideAction()
-                } else {
-                    errorText = Loc.selectAccountError
-                }
+                errorText = Loc.selectAccountError
```
joe-pusya, line 186: You can skip this error, because we have last empty catch with the same `errorText`.
Also we maybe can remove this type of error from `SelectAccountError`

mgolovko: agree

PR #378, opened 2023-09-26
iOS-1732 Add default type to settings and to object creation option

joe-pusya:
1. Add object types selection to the creation and default settings
2. Rename TemplatesSelection module to SetObjectCreationSettings because of object type adding
3. Can't change TemplatesSelection UIKit navigation - try it one more time in separate [task](https://linear.app/anytype/issue/IOS-1790/remake-setobjectcreationsettingscoordinator-to)

https://github.com/anyproto/anytype-swift/assets/6065952/6fc89387-150c-4940-a9c9-3c2c389dff73


mgolovko:
Looks very fresh and stylish

Anytype/Sources/PresentationLayer/ObjectCreationSettings/SetObjectCreationSettingsCoordinator.swift
```diff
@@ -0,0 +1,188 @@
+import UIKit
+import Services
+import SwiftUI
+import AnytypeCore
+
+protocol SetObjectCreationSettingsCoordinatorProtocol: AnyObject {
+    @MainActor
```
mgolovko, line 7: You can mark coordinator class as main actor

Anytype/Sources/ServiceLayer/Object/ObjectType/InstalledObjectTypes/InstalledObjectTypesProvider.swift
```diff
@@ -0,0 +1,67 @@
...
+import Combine
+
+protocol InstalledObjectTypesProviderProtocol: AnyObject {
+    var objectTypes: [ObjectType] { get }
+    var objectTypesPublisher: AnyPublisher<[ObjectType], Never> { get }
+
+    func startSubscription() async
+}
+
+final class InstalledObjectTypesProvider: InstalledObjectTypesProviderProtocol {
```
mgolovko, line 13: it really needed subscription? Maybe use search request
Also we have a ObjectTypeProvider with syncPublisher. This class is very similar

joe-pusya: No, I can't use search request because I need to track object types from the library.
But you are right, I remake logic with ObjectTypeProvider - just filtering types after update, thanks!

mgolovko (approved):

PR #379, opened 2023-09-26
iOS-1591 Copy settings from the active view when create new

joe-pusya:


Anytype/Sources/PresentationLayer/TextEditor/Set/Views/Popups/ViewTypes/SetViewTypesPickerViewModel.swift
```diff
@@ -99,11 +99,10 @@ final class SetViewTypesPickerViewModel: ObservableObject {
     
     private func createView() {
         let name = name.isEmpty ? Loc.SetViewTypesPicker.Settings.Textfield.Placeholder.untitled : name
-        Task {
-            try await dataviewService.createView(
-                DataviewView.created(with: name, type: selectedType),
-                source: source
-            )
+        Task { [weak self] in
+            guard let self else { return }
```
mgolovko, line 103: you capture self immediately when task started for all async calls

mgolovko (approved):

Anytype/Sources/PresentationLayer/TextEditor/Set/Views/Popups/ViewTypes/SetViewTypesPickerViewModel.swift
joe-pusya: you mean we don't nee to make this `guard let self else { return }`?

mgolovko: yep

joe-pusya: Removed in this case

PR #414, opened 2023-10-19
iOS-1679 Spaces | Delete toggles

mgolovko:


Anytype/Sources/PresentationLayer/Settings/FileStorage/onClearCacheSelected.swift
```diff
@@ -0,0 +1,79 @@
+import Foundation
+import Services
+import UIKit
+import Combine
+import AnytypeCore
+
+@MainActor
+final class FileStorageViewModel: ObservableObject {
```
joe-pusya, line 8: file name should be fixed)

joe-pusya (approved):

Anytype/Sources/PresentationLayer/Settings/FileStorage/onClearCacheSelected.swift
mgolovko: lol, fixed)

PR #466, opened 2023-10-27
IOS-1856 Create Space | Fix keyboard glitch for legacy os

mgolovko:


joe-pusya (approved):

Anytype/Sources/PresentationLayer/Modules/SpaceCreate/SpaceCreateView.swift
```diff
@@ -30,5 +30,18 @@ struct SpaceCreateView: View {
...
+private extension View {
+    // Fix glitch when user dismiss screen with opened keyboard
+    @available(iOS, deprecated: 16.4)
+    func ignoreSafeAreaKeyboardLegacy() -> some View {
+        if #available(iOS 16.4, *) {
+            return self
+        } else {
+            return self.ignoresSafeArea(.keyboard)
+        }
     }
```
joe-pusya, line 46: why do we need to make a condition here? `ignoresSafeArea(.keyboard)` works before 16.4 if i'mnot mistaken

mgolovko: From ios 16.4 safe area for keyboard working normally. I disabled keyboard safe area to prevent glitch for ios 16.2

iOS 16.4

https://github.com/anyproto/anytype-swift/assets/2483784/82baa5bf-fb6c-4df4-973f-0714c41a8c76

iOS 16.2 glitch

https://github.com/anyproto/anytype-swift/assets/2483784/5ac5c217-5857-415b-9735-c1c6082fd456

joe-pusya: got it!

PR #470, opened 2023-10-27
IOS-1882 Block Create | Change Analytics

mgolovko:


Anytype/Sources/Analytics/Converters/BlockContentType+Analytics.swift
```diff
@@ -0,0 +1,74 @@
...
+
+extension BlockContentType {
+    var analyticsValue: String {
+        switch self {
+        case .smartblock:
+            return "smartblock"
+        case .text:
+            return "text"
+        case let .file(fileType):
+            return fileType.rawValue
```
mgolovko, line 12: changed here

joe-pusya (approved):

PR #480, opened 2023-10-30
IOS-1929 Renamed Log In button in Log In flow

kmordan:


joe-pusya:
maybe we need to update on `Log In` screen `Login` button too?

mgolovko (approved):

PR #508, opened 2023-11-06
iOS-1909 Make SetObjectCreationCoordinator

joe-pusya:


Anytype/Sources/Models/Extensions/ObjectDetailsExtensions.swift
```diff
@@ -147,4 +147,8 @@ extension BundledRelationsValueProvider {
     var isTemplateType: Bool {
         objectType.uniqueKey == .template
     }
+    
+    var canCreateObject: Bool {
+        setOf.first { $0.isNotEmpty } != nil || isCollection
+    }
```
joe-pusya, line 153: availability plus button condition

mgolovko (approved):

PR #571, opened 2023-11-28
IOS-1883 Profile icon | Fix draw

mgolovko:


Anytype/Sources/PresentationLayer/Common/SwiftUI/IconView/Painters/Extensions/UIImage+Draw.swift
```diff
@@ -8,13 +8,15 @@ extension UIImage {
     }
     
     func drawFit(in bounds: CGRect, maxSize: CGSize) {
+        let multiple = maxSize.width > bounds.width ? 1 - (maxSize.width - bounds.width) / maxSize.width : 1
```
joe-pusya, line 11: If we simplify this equation mathematically, we get this )
`let multiple = maxSize.width > bounds.width ?  bounds.width / maxSize.width : 1`

joe-pusya (approved):

PR #593, opened 2023-12-04
iOS-1293 Core | Mode DataviewService to Services

mgolovko:


Anytype/Sources/PresentationLayer/TextEditor/Set/Models/SetDocument.swift
```diff
@@ -7,7 +7,8 @@ class SetDocument: SetDocumentProtocol {
     let document: BaseDocumentProtocol
     
     var objectId: Services.BlockId { document.objectId }
-    
+    var blockId: BlockId { inlineParameters?.blockId ?? SetConstants.dataviewBlockId }
+    var targetObjectId: BlockId { inlineParameters?.targetObjectID ?? objectId }
```
joe-pusya, line 11: 🔥

joe-pusya (approved):

PR #600, opened 2023-12-06
iOS-1602 Auth | Self-hosted

mgolovko:
https://github.com/anyproto/anytype-swift/assets/2483784/b13580d7-da33-42a0-aa35-4a102aa851cd


Libraryfile
```diff
@@ -1 +1 @@
-MIDDLE_VERSION=v0.29.16
+MIDDLE_VERSION=f76286047
```
mgolovko, line 1: - [ ] Waiting middleware release

PR #605, opened 2023-12-07
IOS-2047 Sharing fixes

bududomasidet:


Modules/Services/Sources/Services/Bookmark/BookmarkService.swift
```diff
@@ -42,7 +42,8 @@ public final class BookmarkService: BookmarkServiceProtocol {
             $0.details = details
             $0.spaceID = spaceId
         }).invoke()
-        return try result.details.toDetails()
+        
+        return ObjectDetails(id: result.objectID)
```
mgolovko, line 46: Why you create empty object details?

mgolovko: Full data needed to open object after create.

PR #634, opened 2023-12-15
IOS-1969 Cant choose exact date in Date relation

bududomasidet:


bududomasidet:
The behaviour could be undefined when there is some tappable button inside SwiftUI's button label.

joe-pusya (approved):

PR #695, opened 2024-01-18
iOS-2109 Base branch | Editing relations

joe-pusya:
https://github.com/anyproto/anytype-swift/assets/6065952/21d541c1-3320-447f-8991-e4e6dcaf8cc5


Anytype/Sources/PresentationLayer/RelationValueFlow/RelationValueCoordinator.swift
```diff
@@ -44,6 +48,24 @@ final class RelationValueCoordinator: RelationValueCoordinatorProtocol,
             return
         }
         
+        if FeatureFlags.newDateRelationCalendarView, case .date(let date) = relation {
```
mgolovko, line 51: Maybe create new relation coordinator view and migrate redesigned relations to swiftui navigation?

joe-pusya: I plan to make it with new feature relations block 👌

PR #716, opened 2024-01-24
iOS-2101 Make new Select relation module pt.1

joe-pusya:
Plan to make this and the next PRs from the previous in editing relation series 

https://github.com/anyproto/anytype-swift/assets/6065952/0b20ccb9-bc25-456b-9a08-3216f85ed70c


Anytype/Sources/PresentationLayer/Common/SwiftUI/RelationDetailsViews/Select/SelectRelationListView.swift
```diff
@@ -0,0 +1,68 @@
+import SwiftUI
+
+struct SelectRelationListView: View {
+    
+    @StateObject var viewModel: SelectRelationListViewModel
+    
+    var body: some View {
+        RelationListContainerView(
```
mgolovko, line 8: Simple common container without module container 🔥

PR #717, opened 2024-01-24
IOS-2132 Migrated SetObjectCreationHelper to async approach

ignatovv:
Migrated SetObjectCreationHelper to async approach

### DIFF
--- a/Anytype/Sources/PresentationLayer/TextEditor/Set/Creation/SetObjectCreationHelper.swift
+++ b/Anytype/Sources/PresentationLayer/TextEditor/Set/Creation/SetObjectCreationHelper.swift
@@ -3,9 +3,8 @@ import AnytypeCore
 
 protocol SetObjectCreationHelperProtocol {
     func createObject(
-        for setDocument: SetDocumentProtocol, setting: ObjectCreationSetting?,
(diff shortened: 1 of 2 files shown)


Anytype/Sources/PresentationLayer/TextEditor/Set/Creation/SetObjectCreationHelper.swift
```diff
@@ -3,9 +3,8 @@ import AnytypeCore
 
 protocol SetObjectCreationHelperProtocol {
     func createObject(
-        for setDocument: SetDocumentProtocol, setting: ObjectCreationSetting?,
-        completion: @escaping ((_ details: ObjectDetails?, _ titleInputType: CreateObjectTitleInputType) -> Void)
-    )
+        for setDocument: SetDocumentProtocol, setting: ObjectCreationSetting?
+    ) async throws -> (ObjectDetails?, CreateObjectTitleInputType)
```
mgolovko, line 7: Tuplets as return type is very difficult to understand results data and send all result data to other function. Tuplets struct should be copied to each function to send it. This is make boilerplate code and difficult refactoring. Try to doesn't using it in future.

PR #721, opened 2024-01-26
IOS-2111 Disable creation of objects in sets by non-page-like types

ignatovv:
Disable creation of objects in sets by non-page-like types

### DIFF
--- a/Anytype/Sources/Models/Extensions/ObjectDetailsExtensions.swift
+++ b/Anytype/Sources/Models/Extensions/ObjectDetailsExtensions.swift
@@ -124,8 +124,16 @@ extension BundledRelationsValueProvider {
         }
     }
     
+    var isGroup: Bool {
+        isSet || isCollection
+    }
+    
     var isCollection: Bool {
         return layoutValue == .collection
+    }
+    
+    var isSet: Bool {
+        return layoutValue == .set
     }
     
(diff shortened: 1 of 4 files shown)


Modules/Services/Sources/Models/Extensions/DetailsLayout+Visible.swift
```diff
@@ -1,6 +1,13 @@
+extension Array {
+    func appended(o: Element) -> [Element] {
+        return self + [o]
+    }
+}
+
```
ignatovv, line 6: I will move it to proper place in next pr

Anytype/Sources/Models/Extensions/ObjectDetailsExtensions.swift
```diff
@@ -152,7 +160,21 @@ extension BundledRelationsValueProvider {
         objectType.isTemplateType
     }
     
-    var canCreateObject: Bool {
-        setOf.first { $0.isNotEmpty } != nil || isCollection
+    func isGroupAndCanCreateObject(setDocument: SetDocumentProtocol) -> Bool {
```
ignatovv, line 163: I prefere to provide setDocument instead of just isRelationsSet: Bool to make it simpler for new api users to understand where to even find this boolean

joe-pusya (approved):

PR #725, opened 2024-01-30
iOS-2215 Notifications | Add base handler

mgolovko:
- Add notification handler
- Add process handler (for feature)
- Add notification coordinator for show notifications

https://github.com/anyproto/anytype-swift/assets/2483784/8563ff4a-6266-48ed-98d8-543e551585a9


Modules/ProtobufMessages/Sources/ServiceMessageHandlerAdapter.swift
```diff
@@ -16,45 +16,49 @@ public protocol ServiceEventsHandlerProtocol: AnyObject {
...
+    public static let shared = ServiceMessageHandlerAdapter()
+    
+    private init() {
+        listen()
     }
     
-    public override init() {
-        super.init()
+    public func addHandler(handler: ServiceEventsHandlerProtocol) {
+        handlers.append(WeakHandler(handler))
```
joe-pusya, line 31: it's a not big deal, but I think we should clear handlers with empty value sometimes (right here for example)

joe-pusya (approved):

Modules/ProtobufMessages/Sources/ServiceMessageHandlerAdapter.swift
mgolovko: agree

PR #728, opened 2024-01-30
IOS-2091 Add support for multi-sectioned types representation

ignatovv:
image


Anytype/Sources/ServiceLayer/Object/Search/SearchService.swift
```diff
@@ -118,6 +122,23 @@ final class SearchService: ObservableObject, SearchServiceProtocol {
         ) { $0.uniqueKey }
     }
     
+    func searchListTypes(text: String, spaceId: String) async throws -> [ObjectDetails] {
```
mgolovko, line 125: What different between searchObjectTypes and this method?

mgolovko (approved):

PR #729, opened 2024-01-31
iOS-2128 Support files as objects

mgolovko:


Modules/Services/Sources/Models/Details/FileDetails.swift
```diff
@@ -0,0 +1,13 @@
...
+    
+    public var name: String { objectDetails.name }
+    public var id: String { objectDetails.id }
+    
+    private let objectDetails: ObjectDetails
+    
+    init(objectDetails: ObjectDetails) {
+        self.objectDetails = objectDetails
+    }
+}
```
joe-pusya, line 13: why we should save objectDetails here instead of setting `name` and `id` in init?

joe-pusya (approved):

mgolovko:
- [ ]  Waiting middleware release

Modules/Services/Sources/Models/Details/FileDetails.swift
mgolovko: fixed

PR #740, opened 2024-02-02
iOS-2243 Default support for graph dataview type

joe-pusya:


Modules/Services/Sources/Models/Block/BlockContent/Content/DataView/DataviewViewType.swift
```diff
@@ -38,6 +41,7 @@ public enum DataviewViewType: Hashable, CaseIterable {
         case .gallery: return "Gallery"
         case .kanban: return "Board"
         case .calendar: return "Calendar"
+        case .graph: return "Graph"
```
mgolovko, line 44: Forgotten about  converter from DataviewTypeEnum to DataviewViewType on 57 line?

mgolovko (approved):

Modules/Services/Sources/Models/Block/BlockContent/Content/DataView/DataviewViewType.swift
joe-pusya: You are right, thanks!

PR #746, opened 2024-02-05
iOS-2184 Sharing extension | support different content

mgolovko:
https://github.com/anyproto/anytype-swift/assets/2483784/397f751b-9b61-45c2-94e0-6c96d9890319


Modules/AnytypeCore/AnytypeCore/Utils/TaskGroup.swift
```diff
@@ -0,0 +1,41 @@
...
+    fileprivate var taskGroup: TaskGroup<SortedItem<ChildTaskResult>>
+    
+    fileprivate init(taskGroup: TaskGroup<SortedItem<ChildTaskResult>>) {
+        self.taskGroup = taskGroup
+    }
+    
+    public mutating func addTask(priority: TaskPriority? = nil, operation: @escaping @Sendable () async -> ChildTaskResult) {
+        let currentIndex = index
+        taskGroup.addTask(priority: priority) { SortedItem(index: currentIndex, content: await operation()) }
+        index+=1
```
joe-pusya, line 21: nit
looks like we don't need `let currentIndex` here, just use `index`
also I really want to add some spaces here `index += 1` ))

joe-pusya (approved):

Modules/AnytypeCore/AnytypeCore/Utils/TaskGroup.swift
mgolovko: needed for capture current value in async task `SortedItem(index: currentIndex`

mgolovko: Spaces fixed

PR #759, opened 2024-02-07
iOS-2156 Add name .asc sort for select / multiselect relation options

joe-pusya:


mgolovko (approved):

joe-pusya:
FYI Product changed his mind and I replace sort with name 👌

ignatovv (approved):

PR #766, opened 2024-02-08
IOS-2255 Type picker | Longtap actions

ignatovv:
image

### DIFF
--- a/Anytype/Sources/ServiceLayer/Object/TypesService/TypesService.swift
+++ b/Anytype/Sources/ServiceLayer/Object/TypesService/TypesService.swift
@@ -1,6 +1,7 @@
 import ProtobufMessages
 import SwiftProtobuf
 import Services
+import AnytypeCore
 
 
 final class TypesService: TypesServiceProtocol {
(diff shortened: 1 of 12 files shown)


Anytype/Sources/ServiceLayer/Object/TypesService/TypesService.swift
```diff
@@ -32,6 +33,17 @@ final class TypesService: TypesServiceProtocol {
         return ObjectType(details: objectDetails)
     }
     
+    func deleteType(_ type: ObjectType, spaceId: String) async throws {
+        guard !type.readonly else {
```
mgolovko, line 37: Is it really needed? Middleware should check it. Also, i think, in feature we can create ObjectService with common object methods like `workspaceObjectListRemove`

mgolovko: You make a limitation. You can delete only ObjectType. If view layer have only id, i can't delete object.

Modules/Services/Sources/Services/Workspace/WorkspaceService.swift
```diff
@@ -34,6 +36,12 @@ public final class WorkspaceService: WorkspaceServiceProtocol {
 		return try ObjectDetails(protobufStruct: result.details)
     }
     
+    public func removeObject(spaceId: String, objectId: String) async throws {
+        try await ClientCommands.workspaceObjectListRemove(.with {
+            $0.objectIds = [objectId]
```
mgolovko, line 41: spaceId needed?

PR #775, opened 2024-02-09
iOS-2122 Fix login error messages

joe-pusya:


Modules/Services/Sources/Services/Auth/Errors/WalletRecoveryError.swift
```diff
@@ -0,0 +1,18 @@
+import Foundation
+import ProtobufMessages
+
+public enum WalletRecoveryError: Error {
+    case badInput
+    case unknownError
+}
+
+extension Anytype_Rpc.Wallet.Recover.Response.Error {
+    var asError: WalletRecoveryError? {
```
mgolovko, line 10: why optional?

joe-pusya: it was optional before, fixed, thanks 👍

mgolovko (approved):

PR #793, opened 2024-02-14
IOS-2274 Common | Add pdf layout & migrate to middleware layout

mgolovko:


Modules/Services/Sources/Models/Details/Entities/DetailsLayout.swift
```diff
@@ -1,27 +1,5 @@
...
-    case collection = 14
-    case audio = 15
-    case video = 16
-    case date = 17
-    case spaceView = 18
-    case participant = 19
-    case database = 20
-    
-    case unknown = -1
-}
```
ignatovv, line 27: 🔥🔥🔥

ignatovv (approved):

PR #841, opened 2024-02-20
iOS-2332 Fix featured relations editing in Sets/Collections

joe-pusya:


Anytype/Sources/PresentationLayer/Common/SwiftUI/Relations/RelationViewProvider/RelationValueView.swift
```diff
@@ -17,7 +17,7 @@ struct RelationValueView: View {
     
     @ViewBuilder
     private func buttonView(with action: (() -> Void)?) -> some View {
-        if action.isNotNil && relation.isEditable {
```
joe-pusya, line 20: we already have this check in relations value coordinator

mgolovko (approved):

PR #843, opened 2024-02-21
IOS-2216 Multiplayer | Update middleware & Space filter

mgolovko:


mgolovko:
Won't fixed on middleware side

PR #845, opened 2024-02-21
IOS-2324 Clipboard | First implementation of paste to create object

ignatovv:
Text only for now
https://github.com/anyproto/anytype-swift/assets/6589348/221a66be-cacd-4b14-ab62-509d36202076


Modules/Services/Sources/Services/BlockService/BlockService.swift
```diff
@@ -26,6 +26,11 @@ public final class BlockService: BlockServiceProtocol {
         return response.blockID
     }
     
+    public func addFirstBlock(contextId: String, info: BlockInformation) async throws -> String {
+        let headerBlockId = "header"
+        return try await add(contextId: contextId, targetId: headerBlockId, info: info, position: .bottom)
```
mgolovko, line 31: We have similar logic in SetObjectCreationHelper. Try to search EditorConstants.headerBlockId

mgolovko (approved):

PR #854, opened 2024-02-22
iOS-2340 Fix active view changing in Sets/Collections

joe-pusya:


Anytype/Sources/PresentationLayer/TextEditor/Set/Models/SetDocument.swift
```diff
@@ -29,15 +29,14 @@ class SetDocument: SetDocumentProtocol {
...
-                }
-            }
+    var blockDataview: BlockDataview? {
+        let blockId = inlineParameters?.blockId ?? SetConstants.dataviewBlockId
+        guard let blockInfo = document.infoContainer.get(id: blockId) else {
+            return nil
+        }
+        if case .dataView(let data) = blockInfo.content {
+           return data
+        } else {
```
joe-pusya, line 39: `document.children` are updated on middle update, but activeViewId has offline logic in Sets, that is why we should get block from the `infoContainer`

ignatovv (approved):

PR #885, opened 2024-02-28
iOS-2108 Remove all uikit relations value views

joe-pusya:


mgolovko:
good-job-1f88a8d26e6fa5efaa7c4ae3cc4b1c85-meme

PR #954, opened 2024-03-08
IOS-2444 Membership | Add ownership info sheet

ignatovv:
image


Anytype/Sources/PresentationLayer/Flows/MembersipCoordinator/MembershipCoordinatorModel.swift
```diff
@@ -8,20 +8,32 @@ final class MembershipCoordinatorModel: ObservableObject {
...
         tierSelectionAssembly: MembershipTierSelectionAssemblyProtocol,
         emailVerificationAssembly: EmailVerificationAssemblyProtocol
     ) {
+        self.membershipService = membershipService
         self.membershipAssembly = membershipAssembly
         self.tierSelectionAssembly = tierSelectionAssembly
         self.emailVerificationAssembly = emailVerificationAssembly
     }
     
+    func onAppear() {
```
joe-pusya, line 31: I think you can just mark onAppear as async here

Anytype/Sources/PresentationLayer/Modules/Membership/TierSelection/MembershipTierSelectionViewModel.swift
```diff
@@ -4,20 +4,29 @@ import Services
...
         membershipService: MembershipServiceProtocol,
         showEmailVerification: @escaping (EmailVerificationData) -> ()
     ) {
-        self.tier = tier
+        self.tierToDisplay = tierToDisplay
         self.membershipService = membershipService
         self.showEmailVerification = showEmailVerification
     }
     
+    func onAppear() {
```
joe-pusya, line 24: and here)

Anytype/Sources/PresentationLayer/Flows/MembersipCoordinator/MembershipCoordinatorModel.swift
ignatovv: membershipService.getStatus() throws
but View modificator .task {} is non throwable
So I have to do catch in view body or wrap call in Task here

joe-pusya: got it!

PR #956, opened 2024-03-08
iOS-2444 Migrate to new DI

ignatovv:


joe-pusya (approved):
👍

Anytype/Sources/PresentationLayer/Flows/MembersipCoordinator/MembershipCoordinator.swift
```diff
@@ -6,11 +6,15 @@ struct MembershipCoordinator: View {
     
     var body: some View {
         model.initialModule()
-            .sheet(item: $model.showTier) {
-                model.tierSelection(tier: $0)
-                    .sheet(item: $model.emailVerificationData) {
-                        model.emailVerification(data: $0)
+            .sheet(item: $model.showTier) { tier in
+                MembershipTierSelectionView(tier: tier) { data in
+                    model.emailVerificationData = data
```
mgolovko, line 11: model.emailVerification is better. Needs to avoid state changing logic in view

PR #1012, opened 2024-03-21
IOS-2531 CI |  Add name for PR checks

mgolovko:


.github/workflows/pr_tests.yml
```diff
@@ -6,6 +6,7 @@ on:
 
 jobs:
   main:
+    name: Test Checks
```
mgolovko, line 9: Will be tested in dev branch. I try to enable required PR checks before merge. For now it doesn't work, because list of available checks in settings is empty

joe-pusya (approved):

PR #1040, opened 2024-03-26
IOS-2521 Buttons | disable buttons if one is active

mgolovko:
https://github.com/anyproto/anytype-swift/assets/2483784/28cf5928-9def-44a0-be82-56ed019839b5

### DIFF
--- a/Anytype/Sources/PresentationLayer/Common/SwiftUI/Buttons/AsyncStandardButtonGroup.swift
+++ b/Anytype/Sources/PresentationLayer/Common/SwiftUI/Buttons/AsyncStandardButtonGroup.swift
@@ -1,0 +1,29 @@
+import Foundation
+import SwiftUI
+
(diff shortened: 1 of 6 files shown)


Anytype/Sources/PresentationLayer/Common/SwiftUI/Buttons/AsyncStandardButtonGroup.swift
```diff
@@ -0,0 +1,29 @@
...
+}
+
+extension EnvironmentValues {
+    var standardButtonGroupDisable: Binding<Bool> {
+        get { self[StandardButtonGroupDisableKey.self] }
+        set { self[StandardButtonGroupDisableKey.self] = newValue }
+    }
+}
+
+struct AsyncStandardButtonGroup<Content: View>: View {
```
mgolovko, line 15: It can be reuse in any view, not just in alert

ignatovv (approved):

PR #1066, opened 2024-03-27
iOS-2526 Analytics | Add extension for file objects if possible

joe-pusya:


Anytype/Sources/Analytics/Converters/FileDetails+Analytics.swift
```diff
@@ -0,0 +1,7 @@
+import Services
+
+extension FileDetails {
+    var analyticsType: AnalyticsObjectType {
+        .file(fileExt: fileExt)
```
mgolovko, line 5: Maybe add lowercase conversion? Extension can be: JPEG, Jpeg, jpeg...

mgolovko (approved):

Anytype/Sources/Analytics/Converters/FileDetails+Analytics.swift
joe-pusya: totally agree, will add!

PR #1078, opened 2024-03-28
IOS-2456 Remove redundant userMembershipPublisher

ignatovv:


joe-pusya (approved):

Anytype/Sources/PresentationLayer/Modules/Membership/InitialScreen/MembershipModuleView.swift
```diff
@@ -7,18 +7,17 @@ struct MembershipModuleView: View {
     @StateObject private var model: MembershipModuleViewModel
     @Environment(\.openURL) private var openURL
     
+    private let membership: MembershipStatus
```
mgolovko, line 10: Why you store state not in model?

PR #1086, opened 2024-04-01
iOS-2162 Set editor | read only

mgolovko:
https://github.com/anyproto/anytype-swift/assets/2483784/ed599569-159b-4692-85e2-b9fde86c5227

### DIFF
--- a/Anytype/Sources/PresentationLayer/Common/SwiftUI/BasicComponents/ButtonForeground.swift
+++ b/Anytype/Sources/PresentationLayer/Common/SwiftUI/BasicComponents/ButtonForeground.swift
@@ -1,0 +1,18 @@
+import Foundation
+import SwiftUI
+
+private struct ButtonForegroundViewModifier: ViewModifier {
+    
+    @Environment(\.isEnabled) private var isEnable
+    
+    func body(content: Content) -> some View {
+        content
+            .foregroundColor(isEnable ? .Button.active : .Button.inactive)
+    }
+}
+
+extension View {
+    func buttonDynamicForegroundColor() -> some View {
+        modifier(ButtonForegroundViewModifier())
+    }
+}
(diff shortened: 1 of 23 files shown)


Anytype/Sources/PresentationLayer/Common/SwiftUI/BasicComponents/ButtonForeground.swift
```diff
@@ -0,0 +1,18 @@
...
+        content
+            .foregroundColor(isEnable ? .Button.active : .Button.inactive)
+    }
+}
+
+extension View {
+    func buttonDynamicForegroundColor() -> some View {
+        modifier(ButtonForegroundViewModifier())
+    }
+}
```
ignatovv, line 18: Naming not really transparent on what to expect from `buttonDynamicForegroundColor`
Also `isEnabe` and `foregroundColor` can be applied not only to buttons but other components as well
I have two optional suggestions
rename `buttonDynamicForegroundColor` to `foregroundColorForDisabledState`
or make it 
dynamicForegroundColor(enabled: Color, disabled: Color)

ignatovv (approved):

Anytype/Sources/PresentationLayer/TextEditor/Set/Views/Popups/ViewPicker/SetViewPicker.swift
```diff
@@ -53,7 +56,8 @@ struct SetViewPicker: View {
         .toolbar {
             ToolbarItem(placement: .navigationBarLeading) {
                 EditButton()
-                    .foregroundColor(Color.Button.active)
+                    .buttonDynamicForegroundColor()
```
joe-pusya, line 59: In relations we just hide buttons in toolbar in non-editable mode, maybe we should make it here in the same way?

joe-pusya (approved):

Anytype/Sources/PresentationLayer/Common/SwiftUI/BasicComponents/ButtonForeground.swift
mgolovko: discussed

Anytype/Sources/PresentationLayer/TextEditor/Set/Views/Popups/ViewPicker/SetViewPicker.swift
mgolovko: fixed

PR #1087, opened 2024-04-01
iOS-2456 Membership | Drop membership cache upon login

ignatovv:


Modules/Services/Sources/Services/Membership/MembershipService.swift
```diff
@@ -32,19 +39,14 @@ final class MembershipService: MembershipServiceProtocol {
     }
     
     public func makeStatusFromMiddlewareModel(membership: MiddlewareMemberhsipStatus) async throws -> MembershipStatus {
-        let cachedTier = try await getTiers(noCache: false).first { $0.type.id == membership.tier }
+        let tier = try await getTiers().first { $0.type.id == membership.tier }
         
-        if let tier = cachedTier {
-            return convertMiddlewareMembership(membership: membership, tier: tier)
+        guard let tier = tier else {
```
joe-pusya, line 44:
```suggestion
        guard let tier else {
```

joe-pusya (approved):

PR #1103, opened 2024-04-02
IOS-2350 Search | Add isHiddenDiscovery filter

mgolovko:


ignatovv (approved):

Modules/Services/Sources/Services/SearchService/SearchHelper.swift
```diff
@@ -313,6 +313,28 @@ public class SearchHelper {
...
+    public static func notHiddenFilters(isArchive: Bool = false, allowHiddenDiscovery: Bool = true) -> [DataviewFilter] {
+        .builder {
+            SearchHelper.isHidden(false)
+            if allowHiddenDiscovery {
+                SearchHelper.isHiddenDiscovery(false)
+            }
+            SearchHelper.isDeletedFilter(isDeleted: false)
+            SearchHelper.isArchivedFilter(isArchived: isArchive)
+        }
+    }
```
joe-pusya, line 336: nit
a little bit confused by naming, maybe `includeHiddenDiscovery` sounds better?

joe-pusya (approved):

Modules/Services/Sources/Services/SearchService/SearchHelper.swift
mgolovko: agree, fixed

PR #1181, opened 2024-04-10
iOS-2651 Resolve require APIs reasons | Update Sentry to the latest version

joe-pusya:


Anytype/Sources/ServiceLayer/Configuration/AnytypeErrorLoggerConfiguration.swift
```diff
@@ -1,15 +1,19 @@
 import Foundation
 import Sentry
 
+@MainActor
 protocol AppErrorLoggerConfigurationProtocol: AnyObject {
-    func setUserId(_ userId: String)
+    func setUserId(_ userId: String) async
```
mgolovko, line 6: Why is this method marked as async and configuration as MainActor ?

joe-pusya: we got warning after sentry update
Screenshot 2024-04-10 at 16 39 30

joe-pusya: remove async

PR #1232, opened 2024-04-16
IOS-2681 Add safari controller wrapper

ignatovv:


Anytype/Sources/PresentationLayer/Modules/Membership/InitialScreen/MembershipModuleView.swift
```diff
@@ -84,13 +86,13 @@ struct MembershipModuleView: View {
...
+                safariUrl = URL(string: AboutApp.pricingLink)!
             }
             MembershipLegalButton(text: Loc.Membership.Legal.privacy) { 
-                openURL(URL(string: AboutApp.privacyPolicyLink)!)
+                safariUrl = URL(string: AboutApp.privacyPolicyLink)!
             }
             MembershipLegalButton(text: Loc.Membership.Legal.terms) { 
-                openURL(URL(string: AboutApp.termsLink)!)
+                safariUrl = URL(string: AboutApp.termsLink)!
             }
```
joe-pusya, line 96: Looks like we don't need force unwrapping here

joe-pusya (approved):
❤️

Anytype/Sources/PresentationLayer/Modules/Membership/InitialScreen/MembershipModuleView.swift
mgolovko: agree. potential crash

PR #1235, opened 2024-04-16
IOS-2681 Support safari sheet and remove UIHelpersDI from auth flow

ignatovv:


joe-pusya (approved):

Anytype/Sources/PresentationLayer/Auth/Auth/AuthView.swift
```diff
@@ -63,7 +65,7 @@ struct AuthView: View {
             .multilineTextAlignment(.center)
             .padding(.horizontal, UIDevice.isPad ? 85 : 38)
             .environment(\.openURL, OpenURLAction { url in
```
mgolovko, line 67: What is it?

mgolovko: You override system openURL?

mgolovko: I not understand goal for this hard decision.
You can read url from view model like `.safariSheet(url: $model.safariUrl, preferredColorScheme: .dark)`

mgolovko (approved):

PR #1275, opened 2024-04-19
IOS-2718 Core | Rename debug project

mgolovko:
If you build release and dev for simulator or device by cabel, difficult to understand where is which build. This change fixed it
Simulator Screenshot - iPhone 15 Pro (member) - 2024-04-19 at 14 00 57

### DIFF
--- a/fastlane/FastfileDeploy
+++ b/fastlane/FastfileDeploy
@@ -95,7 +95,6 @@ platform :ios do
       )
     end
 
-    build_name = options[:release] ? "Anytype" : "Anytype Dev"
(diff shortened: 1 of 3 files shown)


fastlane/FastfileDeploy
```diff
@@ -95,7 +95,6 @@ platform :ios do
       )
     end
 
-    build_name = options[:release] ? "Anytype" : "Anytype Dev"
```
mgolovko, line 98: unused

joe-pusya (approved):
👍

PR #1294, opened 2024-04-22
IOS-2724 Add user property tier to analytics

ignatovv:
Have to make membershipStatusStorage singleton and use it as a source of truth for analytics


Anytype/Sources/Analytics/AnytypeAnalytics/AnytypeAnalytics.swift
```diff
@@ -59,6 +54,10 @@ final class AnytypeAnalytics: AnytypeAnalyticsProtocol {
         userProperties[Keys.networkId] = networkId
     }
     
+    func setMembershipTier(tier: MembershipTier?) {
+        userProperties[Keys.tier] = tier?.name
+    }
```
joe-pusya, line 59: nit
can add String here and avoid `import Service`

joe-pusya (approved):

PR #1348, opened 2024-04-26
iOS-2759 DI | Migrate NewRelationView pt 2

joe-pusya:


Anytype/Sources/PresentationLayer/AddNewRelation flow/CreateNewRelation/Coordinator/NewRelationCoordinatorView.swift
```diff
@@ -21,5 +21,25 @@ struct NewRelationCoordinatorView: View {
                 onFormatSelect: $0.onSelect
             )
         }
+        .sheet(item: $model.newSearchData) {
+            // TODO: Migrate from NewSearchView
+            NewSearchView(
```
mgolovko, line 26: You can create a view for this

mgolovko (approved):

Anytype/Sources/PresentationLayer/AddNewRelation flow/CreateNewRelation/Coordinator/NewRelationCoordinatorView.swift
joe-pusya: ok, will do it 👌

PR #1356, opened 2024-04-29
iOS-2783 Migrate textRelationEditing module

ignatovv:


Anytype/Sources/PresentationLayer/Common/SwiftUI/RelationDetailsViews/Text/TextRelationEditingView.swift
```diff
@@ -142,11 +146,12 @@ struct TextRelationEditingView: View {
 
 #Preview {
     TextRelationEditingView(
-        viewModel: TextRelationEditingViewModel(
+        data: .init(
```
joe-pusya, line 149: discussed not to use `.init` instead of full struct/object name

joe-pusya (approved):

PR #1378, opened 2024-05-03
iOS-2806 Integrate middle v0.34.0-rc1

joe-pusya:


Modules/Services/Sources/Models/Block/BlockConverters/AlignmentConverters.swift
```diff
@@ -5,7 +5,7 @@ import ProtobufMessages
 public extension Anytype_Model_Block.Align {
     var asBlockModel: LayoutAlignment? {
         switch self {
-        case .left: return .left
+        case .left, .justify: return .left
```
joe-pusya, line 8: fallback for left for now 
https://linear.app/anytype/issue/IOS-2804/support-justify-text-align

ignatovv (approved):

PR #1383, opened 2024-05-06
IOS-2814 Cleanup service locator from unused methods

ignatovv:


Anytype.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
```diff
@@ -102,7 +102,7 @@
     {
       "identity" : "swift-collections",
       "kind" : "remoteSourceControl",
-      "location" : "https://github.com/apple/swift-collections",
+      "location" : "https://github.com/apple/swift-collections.git",
```
joe-pusya, line 105: do we need this?

joe-pusya (approved):

PR #1394, opened 2024-05-07
IOS-2818 Membership | Support fallback web based payments

ignatovv:


Modules/Services/Sources/Services/Membership/MembershipModelBuilder.swift
```diff
@@ -0,0 +1,8 @@
+//
+//  File.swift
+//  
+//
+//  Created by Vladimir Ignatov on 07.05.2024.
+//
+
+import Foundation
```
joe-pusya, line 8: do you need this class?

joe-pusya (approved):

Modules/Services/Sources/Services/Membership/MembershipModelBuilder.swift
joe-pusya: oh, I see in the next PR)

PR #1442, opened 2024-05-15
iOS-2859 Core | DI | Delete url opener

mgolovko:


Anytype/Sources/PresentationLayer/Flows/EditorPageFlow/EditorPageCoordinatorView.swift
```diff
@@ -46,5 +46,6 @@ struct EditorPageCoordinatorView: View {
                 UndoRedoView(objectId: $0.value)
             }
             .snackbar(toastBarData: $model.toastBarData)
+            .openUrl(url: $model.openUrlData)
```
mgolovko, line 49: Editor can contains https lins and deeplinks (like anytype://...). `openUrl` can open both links, as legacy url opener

PR #1448, opened 2024-05-17
IOS-2053 Editor | Fix numeric list

mgolovko:
Middleware uses div block for split a lot of child items. If block contains ~40 items, middleware will split one big block to two div blocks with ~20 child items. We did not support this case.

https://github.com/anyproto/anytype-swift/assets/2483784/e35ed9fb-1997-4638-9def-5c756d7ee1bc


joe-pusya (approved):
😨

PR #1524, opened 2024-05-27
IOS-2918 Forbid non-createable types from selection as defaut

ignatovv:


joe-pusya (approved):

Anytype/Sources/Models/Extensions/ObjectType+Extensions.swift
```diff
@@ -16,7 +16,7 @@ extension ObjectType {
         spaceId: "",
         uniqueKey: .empty,
         defaultTemplateId: "",
-        canCreateObjectOfThisType: false,
+        canCreateObjectOfThisType: true,
```
mgolovko, line 19: why?

mgolovko (approved):

PR #1601, opened 2024-06-11
iOS-2968 Call blockDataviewViewSetActive to save view in Sets/Collections

joe-pusya:


Anytype/Sources/PresentationLayer/TextEditor/Set/Views/Popups/ViewPicker/SetViewPicker.swift
```diff
@@ -19,6 +19,7 @@ struct SetViewPicker: View {
         .task {
             await viewModel.startSyncTask()
         }
+        .onChange(of: viewModel.dismiss) { _ in dismiss() }
```
ignatovv, line 22: I find it a bit confusing when properties in views and view models have the same name but different types and uses. Maybe we could rename the view model property to something like shouldDismiss, or even better, pass a dismiss closure directly to the view model and call it there. What do you think?

ignatovv (approved):

Anytype/Sources/PresentationLayer/TextEditor/Set/Views/Popups/ViewPicker/SetViewPicker.swift
joe-pusya: Agree with renaming 🙌

PR #1618, opened 2024-06-13
iOS-2820 Support applying the default type / template during "Turn Into Object" action

joe-pusya:


Anytype/Sources/ServiceLayer/Block/BlockActionHandler/BlockActionService/BlockActionService.swift
```diff
@@ -95,13 +95,14 @@ final class BlockActionService: BlockActionServiceProtocol {
     }
     
     func turnIntoPage(blockId: String, spaceId: String) async throws -> String? {
```
ignatovv, line 97: turnIntoPage -> turnIntoObject

ignatovv (approved):

Anytype/Sources/ServiceLayer/Block/BlockActionHandler/BlockActionService/BlockActionService.swift
joe-pusya: Done

PR #1636, opened 2024-06-18
IOS-2972 Widget | Copy SetObjectWidgetInternalViewModel

mgolovko:


Anytype/Sources/PresentationLayer/Modules/HomeWidgets/Widgets/SpecificInternalModels/SetObjectWidgetLegacyInternalViewModel.swift
```diff
@@ -0,0 +1,203 @@
+import Foundation
+import Services
+import Combine
+import UIKit
+
+@MainActor
+final class SetObjectWidgetLegacyInternalViewModel: ObservableObject, WidgetDataviewInternalViewModelProtocol {
```
mgolovko, line 7: This is copy SetObjectWidgetInternalViewModel.
New copy will be adopt for new widget. This is will be delete in future

joe-pusya (approved):

PR #1646, opened 2024-06-18
IOS-3020 IconView | Support change checkbox for todo icon

mgolovko:
- Enabled by default
- Disable for relations, search, mentions


joe-pusya (approved):
🔥 🔥 🔥

PR #1648, opened 2024-06-18
IOS-3022 Fix some warnings

mgolovko:


Anytype/Sources/Models/Documents/Document/Relation+DependedObject.swift
```diff
@@ -5,29 +5,29 @@ extension Relation {
...
             return []
         case .tag(let tag):
             return tag.selectedTags.map(\.id)
         case .file(let file):
             return file.files.map(\.id)
-        case .unknown(let unknown):
+        case .unknown:
             return []
         }
     }
```
joe-pusya, line 33: maybe join `return []` options?

joe-pusya (approved):

Anytype/Sources/Models/Documents/Document/Relation+DependedObject.swift
mgolovko: yep

PR #1655, opened 2024-06-19
IOS-3017 Remove system errors from validation state

ignatovv:


Anytype/Sources/PresentationLayer/Modules/Membership/Views/NameValidation/MembershipNameValidationViewModel.swift
```diff
@@ -61,7 +61,7 @@ final class MembershipNameValidationViewModel: ObservableObject {
             do {
                 try await membershipService.validateName(name: name, tierType: tier.type)
                 state = .validated
-            } catch let error {
+            } catch let error as MembershipServiceProtocol.ValidateNameError {
```
mgolovko, line 64: I think is better for ignore cancellation error, because middleware can return a lot of different error and user should be visible any error. Alwe we will add a timeout error. It should be visible for user

joe-pusya (approved):

mgolovko (approved):

PR #1658, opened 2024-06-19
IOS-1159 Bind set description appearance to featured relation

ignatovv:


Anytype/Sources/PresentationLayer/TextEditor/Set/Views/Header/SetFullHeader.swift
```diff
@@ -70,13 +70,10 @@ struct SetFullHeader: View {
...
-                AnytypeText(
-                    description,
-                    style: .relation1Regular
-                )
-                .foregroundColor(.Text.primary)
-                .fixedSize(horizontal: false, vertical: true)
+            if model.showDescription {
+                AnytypeText(model.details?.description, style: .relation1Regular)
+                    .foregroundColor(.Text.primary)
+                    .fixedSize(horizontal: false, vertical: true)
```
joe-pusya, line 76: what is description featured and empty? maybe we should add some placeholder or skip showing in this case?

ignatovv: Placeholder will not work because it is readonly label. 
I also do not want to hide it, because it may be misleading. 

I think the best option to keep it as straightforward as possible and look for users/team feedback. I bet no one will report on this one.

joe-pusya (approved):

PR #1666, opened 2024-06-20
iOS-3019 Gallery widget | add cover

mgolovko:
https://github.com/anyproto/anytype-swift/assets/2483784/6294b7ff-b355-4187-88d1-c35e18d17bdd


joe-pusya (approved):
👍

PR #1672, opened 2024-06-21
IOS-3034 Widgets | Add one view for empty state. Use it in gallery

mgolovko:


Anytype/Sources/PresentationLayer/Modules/HomeWidgets/Widgets/Common/WidgetContainerWithEmptyState.swift
```diff
@@ -0,0 +1,28 @@
...
+                WidgetEmptyView(title: Loc.Widgets.Empty.title)
+                    .frame(height: showEmpty ? 72 : 0)
+                Spacer.fixedHeight(8)
+            }
+            .setZeroOpacity(!showEmpty)
+            .transition(.opacity)
+            
+            content
+                .setZeroOpacity(showEmpty)
+        }
```
ignatovv, line 26: Why did you not use `if {} else {}` and use ZStack instead?

ignatovv (approved):

Anytype/Sources/PresentationLayer/Modules/HomeWidgets/Widgets/Common/WidgetContainerWithEmptyState.swift
mgolovko: Not needed. Removed. ZStack was used in old implementation, very often

PR #1673, opened 2024-06-21
IOS-2979 Fix concurrency warning and enable StrictConcurrency for Deeplinks module

ignatovv:


Modules/DeepLinks/Package.swift
```diff
@@ -16,9 +16,14 @@ let package = Package(
     targets: [
         .target(
             name: "DeepLinks",
-            dependencies: []),
+            dependencies: [],
+            swiftSettings: [
+                .enableExperimentalFeature("StrictConcurrency")
```
mgolovko, line 21: 🔥

mgolovko (approved):

PR #1687, opened 2024-06-24
IOS-3046 Widgets | Add button

mgolovko:
Simulator Screenshot - iPhone 15 Pro (owner) - 2024-06-24 at 18 20 58

### DIFF
--- a/Anytype/Sources/PresentationLayer/Common/SwiftUI/EqualFitWidthHStack.swift
+++ b/Anytype/Sources/PresentationLayer/Common/SwiftUI/EqualFitWidthHStack.swift
@@ -1,0 +1,69 @@
+import Foundation
+import SwiftUI
+
+struct EqualFitWidthHStack: Layout {
+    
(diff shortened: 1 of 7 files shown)


Anytype/Sources/PresentationLayer/Common/SwiftUI/EqualFitWidthHStack.swift
```diff
@@ -0,0 +1,69 @@
+import Foundation
+import SwiftUI
+
+struct EqualFitWidthHStack: Layout {
```
ignatovv, line 4: **EqualWidthHStack** looks like shorter name with the same meaning

mgolovko: Fill variant can be implemented without custom layout. I wanted to show that this just only for fit mode.

joe-pusya (approved):

PR #1717, opened 2024-06-27
iOS-3054 Widgets | Update alert

mgolovko:
https://github.com/anyproto/anytype-swift/assets/2483784/d5900c99-43c4-41a2-a1d4-c17d468250da


Anytype/Sources/ServiceLayer/AppVersionService/AppVersionService.swift
```diff
@@ -0,0 +1,56 @@
...
+        
+        let result = try decoder.decode(LoolupResponse.self, from: data)
+        
+        guard let newAppVersion = result.results.first?.version else { return }
+        
+        needsUpdate = appVersion.versionCompare(newAppVersion) == .orderedAscending
+    }
+    
+    private func currentVersion() -> String? {
+        if UserDefaultsConfig.currentVersionOverride.isNotEmpty {
```
mgolovko, line 50: Can be override for test

Anytype/Sources/ServiceLayer/AppVersionService/AppVersionService.swift
```diff
@@ -0,0 +1,56 @@
...
+    }
+    
+    // MARK: - Private
+    
+    private func updateData() async throws {
+        vesionChecked = true
+        
+        guard let appVersion = currentVersion() else { return }
+        
+        guard let url = URL(string: "https://itunes.apple.com/lookup?bundleId=\(AppId.production)") else { return }
```
mgolovko, line 37: For all builds we get production version

Anytype/Sources/PresentationLayer/Modules/HomeWidgets/Container/Submodule/HomeUpdateView.swift
```diff
@@ -0,0 +1,52 @@
...
+                .frame(maxWidth: .infinity)
+            Image(asset: .X18.updateApp)
+                .foregroundColor(.Text.primary)
+                .padding(.horizontal, 16)
+        }
+        .onAppear {
+            withAnimation(.easeInOut(duration: 4).delay(0.5)) {
+                locationState = 6
+            }
+        }
```
ignatovv, line 24: The rationale behind initializing the location state with the value of 6 is rather obscure. Might we consider utilizing boolean or enumerated values for the client-side API, while employing numerical representations internally?

You can find example in `struct ConfettiOverlay: View {`

Also I Suggest to store this animation `.easeInOut(duration: 4).delay(0.5)` in extensions for ease of further reuse.

Anytype/Sources/ServiceLayer/AppVersionService/AppVersionService.swift
ignatovv: Why dont you use #IF Debug for different app ids?

ignatovv (approved):
Animation is super horny 🔥

Anytype/Sources/ServiceLayer/AppVersionService/AppVersionModels.swift
```diff
@@ -0,0 +1,9 @@
+import Foundation
+
+struct LoolupResponse: Codable {
```
joe-pusya, line 3: LookupResponse?

joe-pusya (approved):

Anytype/Sources/PresentationLayer/Modules/HomeWidgets/Container/Submodule/HomeUpdateView.swift
mgolovko: I can try to change 6 to 2 that will be means - two full cycle. 0...1 is one full cycle like percent range.
animation used only in this case. I don't sure it will be reuse anywhere

Anytype/Sources/ServiceLayer/AppVersionService/AppVersionModels.swift
mgolovko: yep

PR #1723, opened 2024-06-28
IOS-3068 Update visual style of sync status

ignatovv:
image


Anytype/Sources/PresentationLayer/TextEditor/EditorPage/Views/Navigation/SyncStatusData.swift
```diff
@@ -6,37 +6,16 @@ struct SyncStatusData {
...
-            return syncedDescription
-        case .failed:
-            return Loc.failedToSyncTryingAgain
+        case .unknown, .offline, .syncing, .synced, .failed:
+            return ""
         case .incompatibleVersion, .UNRECOGNIZED:
-            return Loc.Sync.Status.Version.Outdated.description
+            return Loc.incompatibleVersion
         }
     }
```
joe-pusya, line 20: if we remove description for now, maybe we should remove strings?

joe-pusya (approved):

Anytype/Sources/PresentationLayer/TextEditor/EditorPage/Views/Navigation/SyncStatusData.swift
ignatovv: I am saving them for now. There will be a new info menu, and they will be reused there.

PR #1745, opened 2024-07-04
IOS-3030 Use global singleton for sync status storage

ignatovv:


Anytype/Sources/ServiceLayer/SyncStatus/SyncStatusStorage.swift
```diff
@@ -1,31 +1,34 @@
...
     
+    private var defaultValues = [String: SyncStatus]()
+    
     nonisolated init() { }
     
-    func startSubscription(spaceId: String) {
-        self.spaceId = spaceId
-        
+    func statusPublisher(spaceId: String) -> AnyPublisher<SyncStatus, Never> {
+        updatePublisher
```
mgolovko, line 24: Is not difficult code? You can mark `defaultValues` as publisher and some screens can be subscribe for all states. And `statusPublisher` impl it will be easy

mgolovko (approved):

PR #1768, opened 2024-07-10
iOS-3124 Base document | fix crash with subscriptions array

mgolovko:


Anytype/Sources/Models/Documents/Document/BaseDocument.swift
```diff
@@ -214,19 +214,21 @@ final class BaseDocument: BaseDocumentProtocol {
...
     
-    private func setupSubscriptions() async {
+    @MainActor
+    private func setupSubscriptions() {
         syncStatusStorage.statusPublisher(spaceId: spaceId).sink { [weak self] syncStatus in
             self?.syncStatusSubject.send(syncStatus)
         }.store(in: &subscriptions)
         
-        await accountParticipantsStorage.canEditPublisher(spaceId: spaceId).sink { [weak self] canEdit in
+        accountParticipantsStorage.canEditPublisher(spaceId: spaceId).sink { [weak self] canEdit in
```
mgolovko, line 231: Theory. I couldn't reproduce it.
This `await` call changed thread and make a race for `subscriptions` array.

mgolovko: `.store(in: &subscriptions)` 235 line can be call in different thread

mgolovko: Base document should be actor or mainactor)

ignatovv (approved):

PR #1769, opened 2024-07-10
Ios-3077 Sync status | Show sync status info popup upon click on sync status icon

ignatovv:


Anytype/Sources/FrameworkExtensions/SwiftUI/String+Identifiable.swift
```diff
@@ -0,0 +1,3 @@
+extension String: Identifiable {
```
mgolovko, line 1: We have StringIdentifiable struct

mgolovko (approved):

Anytype/Sources/FrameworkExtensions/SwiftUI/String+Identifiable.swift
ignatovv: I will remove it in upcoming pr

PR #1788, opened 2024-07-15
iOS-3102 Fix file extension conditions

joe-pusya:


Modules/Services/Sources/Models/Details/BundledRelationValueProvider+CustomFields.swift
```diff
@@ -49,7 +49,9 @@ extension BundledRelationsValueProvider {
         if layoutValue == .note {
             title = snippet
         } else if DetailsLayout.fileLayouts.contains(layoutValue) {
-            title = "\(name).\(fileExt)"
+            let ext = ".\(fileExt)"
+            let fileSuffix = fileExt.isNotEmpty && !name.hasSuffix(ext) ? ext : ""
```
mgolovko, line 53: We have the same logic in `FileDetails.fileName`. Needs the sync there?

mgolovko (approved):

Modules/Services/Sources/Models/Details/BundledRelationValueProvider+CustomFields.swift
joe-pusya: Didn't know that, thanks! Unify this logic in FileDetails 🙌

joe-pusya: If everything is ok, you can merge this PR

PR #1794, opened 2024-07-16
IOS-3105 Add background to sync status

ignatovv:


Anytype/Sources/PresentationLayer/TextEditor/EditorPage/Views/NavigationBar/SyncStatus/EditorSyncStatusItem.swift
```diff
@@ -41,16 +50,29 @@ final class EditorSyncStatusItem: UIView {
...
         addSubview(button) { $0.pinToSuperview() }
     }
     
+    private func updateBackgroundColor() {
+        guard let itemState else {
+            backgroundView.backgroundColor = .clear
+            return
+        }
+        
+        backgroundView.backgroundColor = itemState.haveBackground ? .black.withAlphaComponent(0.15) : .clear
```
joe-pusya, line 73: you can move `itemState.haveBackground` to guard, I think it would be clearer

joe-pusya (approved):

PR #1801, opened 2024-07-16
iOS-3158 Version history | Add History menu to object

joe-pusya:

### DIFF
--- a/Anytype/Sources/PresentationLayer/TextEditor/EditorPage/Views/Settings/ObjectSettings/Model/ObjectSettingBuilder.swift
+++ b/Anytype/Sources/PresentationLayer/TextEditor/EditorPage/Views/Settings/ObjectSettings/Model/ObjectSettingBuilder.swift
@@ -1,4 +1,5 @@
 import Services
+import AnytypeCore
 
 final class ObjectSettingBuilder {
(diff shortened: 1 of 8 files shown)


Anytype/Sources/PresentationLayer/TextEditor/EditorPage/Views/Settings/ObjectSettings/Model/ObjectSettingBuilder.swift
```diff
@@ -16,6 +17,10 @@ final class ObjectSettingBuilder {
             }
             
             ObjectSetting.relations
+            
+            if FeatureFlags.versionHistory {
```
mgolovko, line 21: Participant type has a history. It's possible that if the user is a reader, they also don't have a history. You need to use permissions to display it.

mgolovko (approved):

Anytype/Sources/PresentationLayer/TextEditor/EditorPage/Views/Settings/ObjectSettings/Model/ObjectSettingBuilder.swift
joe-pusya: Yes, these are just sketches for sure

PR #1845, opened 2024-07-25
IOS-3234 Message Input | Add placeholder for empty state

mgolovko:
https://github.com/user-attachments/assets/3e2b90bb-38ab-4b0e-8150-057f19ac62eb

### DIFF
--- a/Anytype/Sources/Helpers/Extensions/AttributedString+Extensions.swift
+++ b/Anytype/Sources/Helpers/Extensions/AttributedString+Extensions.swift
@@ -1,0 +1,7 @@
+import Foundation
+
+extension AttributedString {
+    var isEmpty: Bool {
+        startIndex == endIndex
+    }
+}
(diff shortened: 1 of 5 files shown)


Anytype/Sources/Helpers/Extensions/AttributedString+Extensions.swift
```diff
@@ -0,0 +1,7 @@
+import Foundation
+
+extension AttributedString {
+    var isEmpty: Bool {
+        startIndex == endIndex
+    }
+}
```
joe-pusya, line 7: We also can check emptiness like this `characters.isEmpty`, in my opinion indexes look a little bit strange here, but maybe I missed something

joe-pusya (approved):

Anytype/Sources/Helpers/Extensions/AttributedString+Extensions.swift
mgolovko: fixed

PR #2018, opened 2024-08-30
iOS-3334 Add object to collection when dnd between link and collection

joe-pusya:
https://github.com/user-attachments/assets/5e4bfdfa-d893-4198-9cdc-0fac016559a3


ignatovv (approved):
lgtm

PR #2030, opened 2024-09-04
iOS-3365 Chat Message | integrate reactions API (chat branch)

mgolovko:
https://github.com/user-attachments/assets/f5526223-5f17-44a4-a0b6-b1b1dd4aeeb2


Anytype/Sources/PresentationLayer/Modules/Discussion/Subviews/Message/MessageView.swift
```diff
@@ -32,6 +34,9 @@ private struct MessageInternalView: View {
             trailingView
         }
         .padding(.horizontal, 8)
+        .onChange(of: data) {
+            model.update(data: $0)
+        }
     }
```
joe-pusya, line 40: why not published data?

joe-pusya (approved):

PR #2038, opened 2024-09-09
iOS-3377 Show toast on completion when moving blocks

joe-pusya:


ignatovv (approved):

mgolovko (approved):

Anytype/Sources/PresentationLayer/TextEditor/EditorPage/EditorPageBlocksStateManager.swift
```diff
@@ -291,12 +291,19 @@ final class EditorPageBlocksStateManager: EditorPageBlocksStateManagerProtocol {
                     guard let self, let id = targetDocument.children.last?.id,
                           let details = targetDocument.details else { return }
                     if !details.isList {
-                        try await move(position: .bottom, targetId: targetDocument.objectId, dropTargetId: id, blocksIds: filteredBlocksIds)
-                        toastPresenter.showObjectCompositeAlert(
```
mgolovko, line 295: why `try await` syntaxis doesn't work? `toastPresenter` was called after move. `completion` called at the same time 🤔

joe-pusya: it works, but there are some condition inside `move` that can go to return after check and we don't need to call `completion`

mgolovko: Maybe add result reason, to use swift concurrency sintaxis? I expect the completion to be call every time

mgolovko: or add specific error

joe-pusya: I will add result 🙌

PR #2044, opened 2024-09-10
IOS-3387 Change title of entry space

ignatovv:


Anytype/Sources/PresentationLayer/Auth/JoinFlow/VaultView/VaultViewModel.swift
```diff
@@ -45,7 +45,7 @@ final class VaultViewModel: ObservableObject {
             do {
                 state.mnemonic = try await authService.createWallet()
                 let account = try await authService.createAccount(
-                    name: state.soul,
+                    name: Loc.myFirstSpace,
```
joe-pusya, line 48: we use `state.soul` on the last screen of onboarding, should it be prefilled with `Loc.myFirstSpace`? I think so

PR #2053, opened 2024-09-10
iOS-3377 Restrict move to the same object

joe-pusya:
Not only the same block

### DIFF
--- a/Anytype/Sources/PresentationLayer/TextEditor/EditorPage/EditorPageBlocksStateManager.swift
+++ b/Anytype/Sources/PresentationLayer/TextEditor/EditorPage/EditorPageBlocksStateManager.swift
@@ -283,8 +283,16 @@ final class EditorPageBlocksStateManager: EditorPageBlocksStateManagerProtocol {
             if let info = document.infoContainer.get(id: blockId),
                case let .link(content) = info.content {
                 
-                let filteredBlocksIds = movingBlocksIds.filter { $0 != blockId }
                 let targetDocument = documentsProvider.document(objectId: content.targetBlockID)
+                let filteredBlocksIds = movingBlocksIds.filter { [weak self] movingBlockId in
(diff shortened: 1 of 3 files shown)


Anytype/Sources/PresentationLayer/TextEditor/EditorPage/EditorPageBlocksStateManager.swift
```diff
@@ -283,8 +283,16 @@ final class EditorPageBlocksStateManager: EditorPageBlocksStateManagerProtocol {
...
-                let filteredBlocksIds = movingBlocksIds.filter { $0 != blockId }
                 let targetDocument = documentsProvider.document(objectId: content.targetBlockID)
+                let filteredBlocksIds = movingBlocksIds.filter { [weak self] movingBlockId in
+                    guard let self else { return false }
+                    if let info = document.infoContainer.get(id: movingBlockId),
+                       case let .link(content) = info.content {
+                        return content.targetBlockID != targetDocument.objectId
+                    } else {
+                        return true
+                    }
```
ignatovv, line 294: Maybe we can make it more obvious what's happening here. If I hadn't read the title of the pull request, I would have a hard time understanding it.
My ideas: separate method, use a constant with robust name for `content.targetBlockID != targetDocument.objectId`

Also `movingBlocksIds.filter { $0 != blockId }` disappeared, is it ok?

ignatovv (approved):

Anytype/Sources/PresentationLayer/TextEditor/EditorPage/EditorPageBlocksStateManager.swift
mgolovko: probably can implement method like `targetObjectIdByLinkFor(widgetBlockId:)`

mgolovko (approved):

PR #2058, opened 2024-09-11
iOS-3411 All Content | Impl AllContentSubscriptionService for all objects

joe-pusya:
https://github.com/user-attachments/assets/1d93c50b-8748-40a9-901f-71c3470d0341

### DIFF
--- a/Anytype/Sources/PresentationLayer/Modules/AllContent/AllContentView.swift
+++ b/Anytype/Sources/PresentationLayer/Modules/AllContent/AllContentView.swift
@@ -9,7 +9,29 @@ struct AllContentView: View {
     }
     
     var body: some View {
-        Text("This is AllContentView")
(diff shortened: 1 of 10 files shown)


Anytype/Sources/PresentationLayer/Modules/AllContent/AllContentView.swift
```diff
@@ -9,7 +9,29 @@ struct AllContentView: View {
...
+        }
+    }
+    
+    private var content: some View {
+        PlainList {
+            ForEach(model.rows, id: \.id) { row in
+                SearchObjectRowView(
+                    viewModel: row,
+                    selectionIndicatorViewModel: nil
+                )
```
joe-pusya, line 30: most likely a temporary solution with `SearchObjectRowView`

mgolovko (approved):

PR #2080, opened 2024-09-16
iOS-3414 Rename /Picture to /Image in slash menu + add titleSynonyms

joe-pusya:
https://github.com/user-attachments/assets/ba8b577d-faec-49ee-913f-d30b7f84ea99


Anytype/Sources/PresentationLayer/TextEditor/AccessoryView/SlashMenu/Models/Actions/SlashActionMedia.swift
```diff
@@ -75,4 +75,13 @@ enum SlashActionMedia: CaseIterable {
...
     }
+    
+    var titleSynonyms: [String] {
+        switch self {
+        case .image:
+            return [Loc.picture]
+        default:
+            return []
+        }
+    }
```
ignatovv, line 86: Let's populate this synonyms for another cases i suggest just to ask chatgpt for synonyms and pick top three

PR #2081, opened 2024-09-16
IOS-3416 Fix color cheme switch

ignatovv:
I have no idea how but initial coordinator's color scheme affects home.


joe-pusya:
Looks like `.preferredColorScheme(`) switch to definite `colorScheme` for all app screens now and do not switch it to current screen `colorScheme`, can you please check this fix right after auth for example (auth is in dark)?

PR #2100, opened 2024-09-20
IOS-3446 Bump build number

ignatovv:
Fix nighly dev build. Version 57 is already released in Testflight


joe-pusya:
looks like you did't bump build version in all places, did you use this command `bundle exec fastlane increment_build`?

PR #2115, opened 2024-09-24
iOS-3478 All Content | Settings menu design update (left sided checks)

joe-pusya:


Anytype/Sources/PresentationLayer/Common/UIKit/Pickers/FilePicker/CommonViews+Pickers+File+UIKit.swift
```diff
@@ -40,8 +40,8 @@ extension UIKitPickerViewController {
     }
 }
 
-// MARK: - Picker
-class Picker: UIViewController {
+// MARK: - AnytypePicker
+class AnytypePicker: UIViewController {
```
joe-pusya, line 44: Rename our custom `Picker` to `AnytypePicker` because of naming conflict with SwiftUI component

mgolovko (approved):

PR #2119, opened 2024-09-24
IOS-3388 Startup | New animation

mgolovko:
Login

https://github.com/user-attachments/assets/41de2080-8cff-4357-ad84-a3c92467a93a

Forever animation

https://github.com/user-attachments/assets/ee5e6c40-4963-4c50-a37c-f0242e39a446


Anytype/Sources/ApplicationLayer/Launch/LaunchCircle.swift
```diff
@@ -0,0 +1,44 @@
...
+import SwiftUI
+
+struct LaunchCircle: View {
+    
+    private enum Constants {
+        static let maxSide: CGFloat = 128
+        static let middleSide: CGFloat = 48
+        static let minSide: CGFloat = 16
+        static let duration: CGFloat = 1.6
+        static let initialDuration: CGFloat = Constants.duration + Constants.duration
```
ignatovv, line 11: Shouldn't we make it relative depending on the device size? On iPad it will be tiny for example?

ignatovv (approved):
🔥

joe-pusya (approved):

Anytype/Sources/ApplicationLayer/Launch/LaunchCircle.swift
mgolovko: interesting idea, I'll think about it with the designer

PR #2124, opened 2024-09-25
iOS-3482 All Content | Save Sort for each space

joe-pusya:


Anytype/Sources/PresentationLayer/Modules/AllContent/AllContentSavedStates/AllContentSavedStatesService.swift
```diff
@@ -0,0 +1,28 @@
+import AnytypeCore
+
+protocol AllContentSavedStatesServiceProtocol: AnyObject {
```
ignatovv, line 3: Consider AllContent~Saved~State~s~**Storage**Protocol

ignatovv (approved):

Anytype/Sources/PresentationLayer/Modules/AllContent/AllContentSavedStates/AllContentSavedStatesService.swift
joe-pusya: Sounds good, will be renamed in the next PR

PR #2172, opened 2024-10-03
Ios-3524 Show public debug menu for users

ignatovv:
https://github.com/user-attachments/assets/ec5bdf64-e8bd-4747-a92b-750406d6f6ac


Anytype/Sources/PresentationLayer/Debug/Flows/PublicDebug/PublicDebugMenuViewModel.swift
```diff
@@ -0,0 +1,98 @@
...
+    @Injected(\.authService)
+    private var authService: any AuthServiceProtocol
+    @Injected(\.applicationStateService)
+    private var applicationStateService: any ApplicationStateServiceProtocol
+    
+    init() {
+        debugRunProfilerData = debugRunProfilerDataStore
+    }
+    
+    func getLocalStoreData() {
```
joe-pusya, line 32: why are most methods not async?

joe-pusya (approved):

Anytype/Sources/PresentationLayer/Debug/Flows/PublicDebug/PublicDebugMenuViewModel.swift
ignatovv: fixed in next pr

PR #2175, opened 2024-10-04
IOS-3524 Run profiler on app start

ignatovv:
https://github.com/user-attachments/assets/26450b7a-12b6-4766-acd0-141d56018e6d


mgolovko (approved):

Anytype/Sources/ApplicationLayer/AppCoordinator/ApplicationCoordinatorViewModel.swift
```diff
@@ -88,6 +91,13 @@ final class ApplicationCoordinatorViewModel: ObservableObject {
         }
     }
     
+    private func runDebugProfilerIfNeeded() {
```
joe-pusya, line 94: as an idea call this profiler from `GlobalServicesConfiguration` on start

joe-pusya (approved):

PR #2182, opened 2024-10-07
iOS-3501 Remove cross space subscriptions (add spaceId for each subscription)

mgolovko:


joe-pusya (approved):
Screenshot 2024-10-07 at 16 05 44

PR #2192, opened 2024-10-08
iOS-3195 Attachments | Add file from phone (to chat branch)

mgolovko:
https://github.com/user-attachments/assets/f9f347ce-a14b-4773-af7c-c305e051c66d


Anytype/Sources/PresentationLayer/Modules/Discussion/DiscussionViewModel.swift
```diff
@@ -316,4 +327,22 @@ final class DiscussionViewModel: ObservableObject, MessageModuleOutput {
             messagesScrollUpdate = .auto
         }
     }
+    
+    private func handleFilePicker(result: Result<[URL], any Error>) {
+        switch result {
+        case .success(let files):
+            files.forEach { file in
+                let gotAccess = file.startAccessingSecurityScopedResource()
+                if !gotAccess { return }
```
joe-pusya, line 336: `guard gotAccess else { return }`

joe-pusya (approved):

PR #2210, opened 2024-10-11
iOS-3550 Move create object popup in Sets / Collections to anytypesheet

joe-pusya:
https://github.com/user-attachments/assets/e09849cc-eff2-4953-82c1-4d05c7205afc


ignatovv (approved):
lgtm

PR #2258, opened 2024-10-23
IOS-3600 Remove list layouts from default object picker

ignatovv:


Anytype/Sources/PresentationLayer/ObjectTypeSearch/ObjectTypeSearchView.swift
```diff
@@ -170,7 +170,11 @@ struct ObjectTypeSearchView: View {
                 }
             }
         }
-        if !data.isDefault && data.type.canCreateObjectOfThisType {
+        
+        let isNotListLayout = data.type.recommendedLayout.flatMap { !DetailsLayout.setLayouts.contains($0) } ?? false
```
joe-pusya, line 174: use can use `isListType`

joe-pusya (approved):

PR #2275, opened 2024-10-24
fix: blocks moved to another object are out of order

### LINKED ISSUE #2274: Blocks moved to another object are out of order
### Have you read a contributing guide?

- [X] I have read CONTRIBUTING.md
- [X] I have searched the existing issues and didn't find any that were similar
- [X] I have considered creating a pull request with fixes instead of a bug report and want to proceed

### Current Behavior

Given I am on an an object and I select multiple blocks, when I move them to another object, then the objects are moved but the order is not always preserved. The more blocks being moved, the more likely they are out of order.

Simulator Screen Recording - iPhone 16 Pro - 2024-10-24 at 23 18 44

### Expected Behavior

The moved blocks should remain in the order that they are found on the source object.

Simulator Screen Recording - iPhone 16 Pro - 2024-10-24 at 23 17 04

### Steps To Reproduce

1. Create two notes. Open the first and add a list of around 10 elements
2. Select all the elements, then move to the second note.
3. Open the second note.

### Environment

```markdown
- OS: iOS 18
- Version: 0.32.3
```

### Anything else?

_No response_

johnxnguyen:
---

- [x] I understand that contributing to this repository will require me to agree with the [CLA](https://github.com/anyproto/.github/blob/main/docs/CLA.md)

---

#### Issue
Moving several blocks to a new page does not always preserve the order.

#### Cause 
The `EditorPageBlocksStateManager` iterates over all selected blocks and moves them one at a time (seemingly synchronously) via the `BlockActionHandler`. However, the action handler internally fires a `Task` to asynchronously move the block via the `BlockService`. Without any synchronization, it is undefined in which order the tasks will run, resulting in the elements being moved out of order.

#### Solution
Don't hide a task inside `BlockActionHandler.moveToPage(blockId:pageId)`, instead make it `async throws`, then await for each move to complete before iterating to the next block.

#### Note
I marked the method as throws so errors are propagated up, but it gets ignored anyway in the enclosing task. I didn't notice error handling in similar methods (most of the errors get lost in the `Task`) so I wasn't sure if or how it should be handled.

### What type of PR is this? (check all applicable)

- [ ] 🍕 Feature
- [x] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] 🔥 Performance Improvements
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI

### Related Tickets & Documents
Fixes #2274

### Mobile & Desktop Screenshots/Recordings

Before:
Simulator Screen Recording - iPhone 16 Pro - 2024-10-24 at 23 18 44

After:
Simulator Screen Recording - iPhone 16 Pro - 2024-10-24 at 23 17 04

### Added tests?

- [ ] 👍 yes
- [ ] 🙅 no, because they aren't needed
- [x] 🙋 no, because I need help

I didn't see any tests for relevant objects (such as `EditorPageBlocksStateManager`) so I'm not sure what your testing strategy is.

### Added to documentation?

- [ ] 📜 README.md
- [ ] 📓 [tech-docs](https://github.com/anyproto/tech-docs)
- [x] 🙅 no documentation needed

### DIFF
--- a/Anytype/Sources/ServiceLayer/Block/BlockActionHandler/BlockActionHandler.swift
+++ b/Anytype/Sources/ServiceLayer/Block/BlockActionHandler/BlockActionHandler.swift
@@ -130,13 +130,12 @@ final class BlockActionHandler: BlockActionHandlerProtocol {
         service.delete(blockIds: blockIds)
     }
     
-    func moveToPage(blockId: String, pageId: String) {
+    func moveToPage(blockId: String, pageId: String) async throws {
         AnytypeAnalytics.instance().logMoveBlock()
-        Task {
-            try await blockService.moveToPage(objectId: document.objectId, blockId: blockId, pageId: pageId)
-        }
+        try await blockService.moveToPage(objectId: document.objectId, blockId: blockId, pageId: pageId)
     }
-    
+
+
     func createEmptyBlock(parentId: String, spaceId: String) {
         Task {
             let emptyBlock = BlockInformation.emptyText
--- a/Anytype/Sources/ServiceLayer/Block/BlockActionHandler/BlockActionHandler.swift
+++ b/Anytype/Sources/ServiceLayer/Block/BlockActionHandler/BlockActionHandler.swift
@@ -130,9 +130,9 @@ final class BlockActionHandler: BlockActionHandlerProtocol {
         service.delete(blockIds: blockIds)
     }
     
-    func moveToPage(blockId: String, pageId: String) async throws {
+    func moveToPage(blockIds: [String], pageId: String) async throws {
         AnytypeAnalytics.instance().logMoveBlock()
-        try await blockService.moveToPage(objectId: document.objectId, blockId: blockId, pageId: pageId)
+        try await blockService.moveToPage(objectId: document.objectId, blockIds: blockIds, pageId: pageId)
     }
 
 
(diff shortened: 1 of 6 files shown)


Anytype/Sources/ServiceLayer/Block/BlockActionHandler/BlockActionHandler.swift
```diff
@@ -130,13 +130,12 @@ final class BlockActionHandler: BlockActionHandlerProtocol {
         service.delete(blockIds: blockIds)
     }
     
-    func moveToPage(blockId: String, pageId: String) {
+    func moveToPage(blockId: String, pageId: String) async throws {
         AnytypeAnalytics.instance().logMoveBlock()
-        Task {
-            try await blockService.moveToPage(objectId: document.objectId, blockId: blockId, pageId: pageId)
-        }
+        try await blockService.moveToPage(objectId: document.objectId, blockId: blockId, pageId: pageId)
```
mgolovko, line 135: moveToPage support multiple block move. Is better for use one call for prevent multiple animaitons in editor

johnxnguyen: Sure sounds good. Just so I understand, are you suggesting that we make the block service capable of moving multiple blocks in a single method call?

johnxnguyen: I see what you meant, the client command already accepts an array of block ids. Block service now accepts an array of block ids (see dca6160e6cc74e6e78aa90fcbcd7e8d902370789), thanks for the tip!

ignatovv (approved):
Good job 🙏

mgolovko:
Your changes have been passed to qa

PR #2300, opened 2024-10-30
IOS-3643 Home Chat |  Update top panel (without blur)

mgolovko:
https://github.com/user-attachments/assets/ade62482-dcb7-4996-8cd2-f893874489ac


ignatovv (approved):
wow, looks cool

PR #2311, opened 2024-10-31
iOS-3653 All content | Add types list to all objects

ignatovv:
image

### DIFF
--- a/Anytype/Sources/PresentationLayer/Modules/AllContent/AllContentViewModel.swift
+++ b/Anytype/Sources/PresentationLayer/Modules/AllContent/AllContentViewModel.swift
@@ -10,7 +10,7 @@ protocol AllContentModuleOutput: AnyObject {
 
 @MainActor
 final class AllContentViewModel: ObservableObject {
-
+    
     private var details = [ObjectDetails]()
     private var objectsToLoad = 0
     var firstOpen = true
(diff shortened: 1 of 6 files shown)


Anytype/Sources/PresentationLayer/Modules/AllContent/AllContentViewModel.swift
```diff
@@ -138,17 +138,23 @@ final class AllContentViewModel: ObservableObject {
...
+            return
+        }
+        
+        let today = Date()
+        let dict = OrderedDictionary(
+            grouping: details,
+            by: { dateFormatter.localizedString(for: sortValue(for: $0) ?? today, relativeTo: today) }
+        )
+        
+        if dict.count == 1 {
```
joe-pusya, line 152: it means when you create the new space - there is no sections there, because of only one section `Today` - is it ok?

joe-pusya (approved):

Anytype/Sources/PresentationLayer/Modules/AllContent/AllContentViewModel.swift
ignatovv: Yep, I think it is intended behaviour. There is no value in Today section if it is the only one

PR #2328, opened 2024-11-04
IOS-3652 Add basic type layout support

ignatovv:
https://github.com/user-attachments/assets/3204cdcb-ea9e-4e7f-be0a-8e590f75c2e8

### DIFF
--- a/Modules/Services/Sources/Models/Extensions/DetailsLayoutExtension.swift
+++ b/Modules/Services/Sources/Models/Extensions/DetailsLayoutExtension.swift
@@ -9,12 +9,13 @@ public extension DetailsLayout {
     static let mediaLayouts: [DetailsLayout] = [ .image, .audio, .video ]
     static let fileAndMediaLayouts = DetailsLayout.fileLayouts + DetailsLayout.mediaLayouts
     
-    static let visibleLayouts: [DetailsLayout] = listLayouts + editorLayouts + [.bookmark] + [.participant] + dateLayout + chatLayout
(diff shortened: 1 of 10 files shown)


joe-pusya (approved):

Modules/Services/Sources/Models/Extensions/DetailsLayoutExtension.swift
```diff
@@ -9,12 +9,13 @@ public extension DetailsLayout {
     static let mediaLayouts: [DetailsLayout] = [ .image, .audio, .video ]
     static let fileAndMediaLayouts = DetailsLayout.fileLayouts + DetailsLayout.mediaLayouts
     
-    static let visibleLayouts: [DetailsLayout] = listLayouts + editorLayouts + [.bookmark] + [.participant] + dateLayout + chatLayout
+    static let visibleLayouts: [DetailsLayout] = listLayouts + editorLayouts + [.bookmark, .participant, .objectType] + dateLayout + chatLayout
```
joe-pusya, line 12: maybe we should add toggle for objectType?

ignatovv: The issue is that we cannot toggle the AllContentType enum. Therefore, in an emergency, it will require manual disabling of this feature, and toggling does not really make sense in this case.

joe-pusya: I mean toggle adding to `visibleLayouts` while the objectType page is not ready

PR #2331, opened 2024-11-04
iOS-3667 Add dynamic Bin widget to home screen

joe-pusya:
https://github.com/user-attachments/assets/8029369d-435a-4210-8fe0-e4076e7faad5

### DIFF
--- a/Anytype/Sources/PresentationLayer/Modules/AllContent/AllContentSettingsMenu.swift
+++ b/Anytype/Sources/PresentationLayer/Modules/AllContent/AllContentSettingsMenu.swift
@@ -1,5 +1,6 @@
 import SwiftUI
 import Services
+import AnytypeCore
 
 struct AllContentSettingsMenu: View {
     @Binding var state: AllContentState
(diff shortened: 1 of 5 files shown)


Anytype/Sources/PresentationLayer/Modules/AllContent/AllContentSettingsMenu.swift
```diff
@@ -20,8 +21,10 @@ struct AllContentSettingsMenu: View {
             mode
             Divider()
             sortByMenu
-            Divider()
-            bin
+            if !FeatureFlags.showBinWidgetIfNotEmpty {
+                Divider()
+                bin
+            }
```
ignatovv, line 27: should we delete bin from settings menu? I think it might be a nice addition.

ignatovv (approved):

Anytype/Sources/PresentationLayer/Modules/AllContent/AllContentSettingsMenu.swift
joe-pusya: yes, maybe, asked Kostya 🙌

PR #2337, opened 2024-11-05
iOS-3652 | Support opening of types

ignatovv:
https://github.com/user-attachments/assets/4c15f36d-441a-4bc7-a074-4e0096935ddb


Anytype/Sources/PresentationLayer/Modules/ObjectType/ObjectTypeView.swift
```diff
@@ -9,13 +9,52 @@ struct ObjectTypeView: View {
...
         VStack {
-            AnytypeText(model.title, style: .title)
-            AnytypeText("TBD", style: .subheading).foregroundColor(.Text.secondary)
+            Rectangle().frame(height: 48).foregroundStyle(Color.Text.secondary) // Navbar placeholder
+            VStack(spacing: 0) {
+                Spacer.fixedHeight(32)
+                header
+                templates
+                Spacer()
+            }.padding(.horizontal, 20)
```
joe-pusya, line 24: I think templates don't need to have horizontal padding - they should scroll after the screen edges)

joe-pusya (approved):

PR #2349, opened 2024-11-08
iOS-3708 Integrate @Entry macro

joe-pusya:
Before

```
struct KeyboardDismiss: EnvironmentKey {
    static let defaultValue: () -> Void = {}
}

extension EnvironmentValues {
    var keyboardDismiss: () -> Void {
        get { self[KeyboardDismiss.self] }
        set { self[KeyboardDismiss.self] = newValue }
    }
}
```

After

```
extension EnvironmentValues {
    @Entry var keyboardDismiss: () -> Void = {}
}

```


ignatovv (approved):
🔥🔥

PR #2375, opened 2024-11-18
IOS-3652 Add delete type action

ignatovv:
https://github.com/user-attachments/assets/56a7e092-e59e-4f37-a360-c812bfc5549b

### DIFF
--- a/Anytype/Sources/PresentationLayer/Modules/ObjectType/ObjectTypeViewModel.swift
+++ b/Anytype/Sources/PresentationLayer/Modules/ObjectType/ObjectTypeViewModel.swift
@@ -12,18 +12,20 @@ final class ObjectTypeViewModel: ObservableObject, RelationsListModuleOutput {
     @Published var syncStatusData: SyncStatusData?
     
     @Published var showSyncStatusInfo = false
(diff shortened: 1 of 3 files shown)


Anytype/Sources/PresentationLayer/Modules/ObjectType/ObjectTypeViewModel.swift
```diff
@@ -95,6 +103,16 @@ final class ObjectTypeViewModel: ObservableObject, RelationsListModuleOutput {
         }
     }
     
+    func onDeleteTap() {
+        showDeleteConfirmation.toggle()
+    }
+    
+    func onDeleteConfirm() async throws {
+        try await objectActionsService.delete(objectIds: [document.objectId])
+        showDeleteConfirmation = false
```
joe-pusya, line 112: nit
you can `dismiss` in `TypeDeleteAlert` and don't handle `showDeleteConfirmation` here

joe-pusya (approved):

joe-pusya:
one more question - should we add this deletion alert on swipe left actions in the All content list?

PR #2408, opened 2024-11-22
iOS-3745 Date as an object | Filter relations

joe-pusya:
And move to `objectShow` - we don't need to call `objectOpen` for virtual objects


Anytype/Sources/PresentationLayer/Modules/Date/DateViewModel.swift
```diff
@@ -152,20 +153,40 @@ final class DateViewModel: ObservableObject {
             document.objectId,
             spaceId: document.spaceId
         )
-        // Set mentions relation first
-        let reorderedRelationsKeys = relationsKeys?.reordered(by: [BundledRelationKey.mentions.rawValue], transform: { $0 }) ?? []
```
joe-pusya, line 156: will be implemented on the middle side

PR #2425, opened 2024-11-26
iOS-3750 Fix typo

joe-pusya:


ignatovv (approved):

Anytype/Generated/Strings.swift
```diff
@@ -1710,7 +1710,7 @@ internal enum Loc {
       }
     }
     internal enum DeleteSharingLink {
-      internal static let message = Loc.tr("Localizable", "SpaceShare.DeleteSharingLink.Message", fallback: "New members won’t be able to join the space. You can generate a new link anytype")
+      internal static let message = Loc.tr("Localizable", "SpaceShare.DeleteSharingLink.Message", fallback: "New members won’t be able to join the space. You can generate a new link anytime")
```
ignatovv, line 1713: internet suggest to use "any time"

ignatovv (approved):

Anytype/Generated/Strings.swift
joe-pusya: Seems like both variables are correct, I'll leave it like this for now 🙌

PR #2436, opened 2024-11-28
IOS-3781 Home | Add swipe between chat and widgets (from iOS 18)

mgolovko:
https://github.com/user-attachments/assets/6166db11-f6f4-4481-bef8-c5427cfebab6


joe-pusya (approved):
🙀 🚀

PR #2445, opened 2024-11-28
IOS-3636 Provided parsed relations api for type objects

ignatovv:


joe-pusya (approved):

Anytype/Sources/Models/Documents/Document/BaseDocument+Extensions.swift
```diff
@@ -76,4 +76,26 @@ extension BaseDocumentProtocol {
             .removeDuplicates()
             .eraseToAnyPublisher()
     }
+    
+    func buildParsedRelationsForType() -> ParsedRelations {
```
mgolovko, line 80: If the client reads this field very frequently, it will cause performance issues. The other parsedRelations are updated via the update diff. I think this field should be included in the design of the BaseDocument update.

mgolovko: You are also using `Container.shared`, which means the document is designed incorrectly for this field.

PR #2462, opened 2024-12-03
IOS-3727 Fixed wrong index crash

ignatovv:


Anytype/Sources/PresentationLayer/TextEditor/EditorPage/Models/BlockViewModelBuilder.swift
```diff
@@ -87,7 +87,7 @@ final class BlockViewModelBuilder {
             items.insert(.system(spacer), at: featureRelationsIndex + 2)
         } else if let featureRelationsIndex  {
             items.insert(.system(spacer), at: featureRelationsIndex + 1)
-        } else if let openFileButton  {
+        } else if let openFileButton, items.count > 0 {
```
joe-pusya, line 90: interesting, how it can be `0` if we have this line above `var items = viewModels.map(EditorItem.block)`?

joe-pusya (approved):

Anytype/Sources/PresentationLayer/TextEditor/EditorPage/Models/BlockViewModelBuilder.swift
ignatovv: Sometimes block subscription return empty list on updates

PR #2592, opened 2024-12-26
iOS-3925 Home | New Navigation | Back button

mgolovko:
https://github.com/user-attachments/assets/826e75a2-1b6e-499f-b65f-dca2051e4d8d


Anytype/Sources/FrameworkExtensions/SwiftUI/TappableArea.swift
```diff
@@ -1,18 +1,43 @@
 import Foundation
 import SwiftUI
 
+struct IncreaseTapGestureV2Midifier: ViewModifier {
```
joe-pusya, line 4: IncreaseTapGestureV2Midifier -> IncreaseTapGestureV2Modifier

mgolovko: fixed

PR #2618, opened 2025-01-10
iOS-3933 SpaceHub | Fix space background performance

mgolovko:
- Fix glitch when open hub with pop gesture

### DIFF
--- a/Anytype/Sources/Design system/Colors/Gradients/CoverGradient.swift
+++ b/Anytype/Sources/Design system/Colors/Gradients/CoverGradient.swift
@@ -1,19 +1,23 @@
 import SwiftUI
 
-struct CoverGradientData: Identifiable {
+struct CoverGradientData: Identifiable, Equatable {
     let name: String
     let startColor: Color
     let endColor: Color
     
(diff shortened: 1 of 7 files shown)


Anytype/Sources/Design system/Colors/Gradients/CoverGradient.swift
```diff
@@ -1,19 +1,23 @@
...
-struct CoverGradientData: Identifiable {
+struct CoverGradientData: Identifiable, Equatable {
     let name: String
     let startColor: Color
     let endColor: Color
     
     var id: String { "\(name)\(startColor)\(endColor)" }
+}
+
+struct CoverGradientView: View, Equatable {
```
joe-pusya, line 11: `Equatable` is needed?

joe-pusya (approved):
👍

Anytype/Sources/Design system/Colors/Gradients/CoverGradient.swift
mgolovko: Yes. For use https://developer.apple.com/documentation/swiftui/equatableview

PR #2627, opened 2025-01-14
IOS-3941 Fallback layout to basic

ignatovv:
Prerequisite for primitives migration


Modules/Services/Sources/Models/Details/BundledRelationValueProvider+CustomFields.swift
```diff
@@ -8,7 +8,7 @@ extension BundledRelationsValueProvider {
     
     public var layoutValue: DetailsLayout {
         guard let number = layout, let layout = DetailsLayout(rawValue: number) else {
-            return .UNRECOGNIZED(layout ?? -1)
+            return .basic
```
mgolovko, line 11: "Set" and other objects (not page types) can be broken if it open as page and try to edit. 
Object icons logic will be broken.

mgolovko (approved):

PR #2673, opened 2025-01-22
IOS-4016 Add inline ResultWithError protocol conformance

ignatovv:


joe-pusya (approved):
Screenshot 2025-01-22 at 13 26 54

PR #2733, opened 2025-02-05
iOS-4072 Chat Message | Experiment | Your bubble color for background

mgolovko:
https://github.com/user-attachments/assets/addf04ab-a3bb-4b62-8139-57f9ac30cd04


joe-pusya (approved):
Wow! Looks cool! ✨

PR #2817, opened 2025-02-18
IOS-4062 | Supported custom icons for objects

ignatovv:
https://github.com/user-attachments/assets/175b3f1f-a947-4baf-9a3d-c807a7d8663b

### DIFF
--- a/Modules/Services/Sources/Models/Details/ObjectIconBuilder.swift
+++ b/Modules/Services/Sources/Models/Details/ObjectIconBuilder.swift
@@ -43,7 +43,7 @@ public final class ObjectIconBuilder: ObjectIconBuilderProtocol {
     
     private func icon(relations: BundledRelationsValueProvider) -> ObjectIcon? {
         switch relations.layoutValue {
-        case .basic, .set, .collection, .image, .objectType, .chat:
+        case .basic, .set, .collection, .image, .chat:
             return basicIcon(iconImage: relations.iconImage, iconEmoji: relations.iconEmoji)
         case .profile, .participant:
             return profileIcon(iconImage: relations.iconImage, objectName: relations.objectName)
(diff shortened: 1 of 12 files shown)


Modules/Services/Sources/Models/Details/ObjectIconBuilder.swift
```diff
@@ -84,4 +86,14 @@ public final class ObjectIconBuilder: ObjectIconBuilderProtocol {
     private func fileIcon(fileMimeType: String, name: String) -> ObjectIcon {
         return .file(mimeType: fileMimeType, name: name)
     }
+    
+    private func objectTypeIcon(iconName: String, iconOption: Int?, iconImage: String, iconEmoji: Emoji?) -> ObjectIcon? {
+        if FeatureFlags.newTypeIcons {
+            if iconName.isNotEmpty, let customIcon = CustomIcon(rawValue: iconName) {
```
mgolovko, line 92: Don't you think CustomIcon is too generic for the object icon name?

ignatovv: I was thinking about it, but frankly speaking, I just did not come up with anything more wholesome. Do you have any ideas off the top of your head?

mgolovko: `ObjectIconAsset` or `[Some]Asset`. Where `Asset` means that from resource
`ObjectIconName` for similar with mw name.

mgolovko (approved):

Modules/Services/Sources/Models/Details/ObjectIconBuilder.swift
mgolovko: We can continue discussing in slack channel

PR #2836, opened 2025-02-20
Ios-4062 New icons | Support appearance of new icons for types

ignatovv:
https://github.com/user-attachments/assets/9252adc5-46cd-4ae4-a2c4-76cf9730a231


Modules/Services/Sources/Models/ObjectType.swift
```diff
@@ -61,7 +85,7 @@ extension ObjectType: DetailsModel {
         self.init(
             id: details.id,
             name: details.name,
-            iconEmoji: details.iconEmoji,
+            typeIcon: ObjectTypeIcon(icon: details.customIcon, iconColor: details.customIconColor, emoji: details.iconEmoji),
```
mgolovko, line 88: What about ObjectIconBuilder? Maybe use `objectIcon` field for sync logic in one place?

ignatovv: Good call 🙏
Fixed

mgolovko (approved):

PR #2841, opened 2025-02-21
iOS-4203 Core | Send console log

mgolovko:


Anytype/Sources/ApplicationLayer/AppSessionTracker/StdOutFileStream.swift
```diff
@@ -0,0 +1,91 @@
...
+        
+        let outPipe = Pipe()
+        dup2(outPipe.fileHandleForWriting.fileDescriptor, STDOUT_FILENO)
+
+        let errPipe = Pipe()
+        dup2(errPipe.fileHandleForWriting.fileDescriptor, STDERR_FILENO)
+        
+        readQueue = DispatchQueue(
+            label: "io.anytype.stdout",
+            qos: .userInteractive,
```
joe-pusya, line 51: is it not too hight qos?

joe-pusya (approved):

Anytype/Sources/ApplicationLayer/AppSessionTracker/StdOutFileStream.swift
mgolovko: agree

PR #2854, opened 2025-02-25
iOS-4227 Add opportunity to set space icon during space creation

joe-pusya:
https://github.com/user-attachments/assets/3c925d02-ab4d-410c-8709-513d040bba87


Anytype/Sources/PresentationLayer/Modules/SpaceCreate/SpaceCreateView.swift
```diff
@@ -51,6 +48,28 @@ struct SpaceCreateView: View {
         }
         .ignoreSafeAreaKeyboardLegacy()
         .background(Color.Background.primary)
+        .onChange(of: model.spaceName) {
+            model.updateNameIconIfNeeded($0)
+        }
+        .sheet(isPresented: $model.showLocalIconPicker) {
```
mgolovko, line 54: Maybe add coordinator for space creation flow?

joe-pusya: Agree, will do)

PR #2878, opened 2025-02-28
IOS-4248 Profile | Show display name & small object icon refactoring 

mgolovko:


Anytype/Sources/Models/Extensions/ObjectDetailsExtensions.swift
```diff
@@ -95,6 +95,10 @@ extension BundledRelationsValueProvider {
     var isSupportedForOpening: Bool { layoutValue.isSupportedForOpening }
     
     var isVisibleLayout: Bool { layoutValue.isVisible }
+    
+    var displayName: String {
```
ignatovv, line 99: It will be nice to use it [here](https://github.com/anyproto/anytype-swift/blob/2c78cafe46df82a471978cedf9019c9d1e75b164/Anytype/Sources/PresentationLayer/SpaceSettings/SpaceSettings/SpaceInfoBuilder/SpaceSettingsInfoBuilder.swift#L39) as well

mgolovko: In those cases the Participant is used, not the ObjectDetails. It won't combine them(

PR #2892, opened 2025-03-03
iOS-4236 Editor | open media picker after create a block

mgolovko:
https://github.com/user-attachments/assets/c197a2da-b17f-40eb-b99b-2e3a580b1e70


ignatovv (approved):
lgtm

PR #2934, opened 2025-03-11
IOS-4302 Primitives | Add resolve conflicts section to object settings

ignatovv:
https://github.com/user-attachments/assets/9c77fd15-093f-41a0-a084-d7850328d214

### DIFF
--- a/Modules/Assets/Sources/Assets/Resources/Assets.xcassets/DesignSystem/x18/Attention.imageset/Contents.json
+++ b/Modules/Assets/Sources/Assets/Resources/Assets.xcassets/DesignSystem/x18/Attention.imageset/Contents.json
@@ -10,6 +10,6 @@
     "version" : 1
   },
   "properties" : {
-    "template-rendering-intent" : "template"
+    "template-rendering-intent" : "original"
   }
 }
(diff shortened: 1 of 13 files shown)


Modules/Assets/Sources/Assets/Resources/Assets.xcassets/DesignSystem/x18/Attention.imageset/Contents.json
```diff
@@ -10,6 +10,6 @@
     "version" : 1
   },
   "properties" : {
-    "template-rendering-intent" : "template"
+    "template-rendering-intent" : "original"
```
joe-pusya, line 13: is it on demand? we usually just set .foregroundColor for Image

joe-pusya (approved):

Modules/Assets/Sources/Assets/Resources/Assets.xcassets/DesignSystem/x18/Attention.imageset/Contents.json
ignatovv: Yep, it is tricky image with background. We use templates for images without background generally. 
image

PR #2935, opened 2025-03-11
IOS-4307 Primitives | Remove relation creation from featured section

ignatovv:


Modules/Services/Sources/Models/Details/BundledRelationValueProvider+CustomFields.swift
```diff
@@ -7,6 +7,7 @@ extension BundledRelationsValueProvider {
     public var isDone: Bool { done }
     
     public var layoutValue: DetailsLayout {
+        // use "resolvedLayout"; "layout" is deprecated
```
mgolovko, line 10: Maybe rename layoutValue to resolvedLayout in all app?

mgolovko (approved):

PR #2938, opened 2025-03-12
IOS-4104 Supported axis for anytype text fields

ignatovv:
https://github.com/user-attachments/assets/c4bcf76f-4ecb-49d4-a4d6-9ff0b0464c66


Anytype/Sources/PresentationLayer/Common/SwiftUI/TextField/AnytypeTextField.swift
```diff
@@ -3,11 +3,19 @@ import SwiftUI
 struct AnytypeTextField: View {
     let placeholder: String
     let font: AnytypeFont
+    let axis: Axis
```
mgolovko, line 6: you can write `var axis: Axis = . horizontal` and init will be generate automatically)
just FYI)

mgolovko (approved):

PR #3014, opened 2025-03-26
iOS-4324 Chat date header | make floating

mgolovko:
https://github.com/user-attachments/assets/95350402-8420-4e92-9199-0e4fdd0a5ad1


joe-pusya (approved):
Fire 🔥

PR #3133, opened 2025-04-14
iOS-4454 Make group for each target

joe-pusya:
```
public struct TargetsConstants {    
    public static var appGroup: String  {
        switch CoreEnvironment.targetType {
        case .debug:
            return "group.io.anytype.app.dev"
        case .releaseAnytype:
            return "group.io.anytype.app"
        case .releaseAnyApp:
            return "group.org.any.app"
        }
    }
}
```


Modules/AnytypeCore/AnytypeCore/Utils/TargetsConstants.swift
```diff
@@ -0,0 +1,9 @@
+public enum TargetsConstants {
+#if DEBUG || RELEASE_NIGHTLY
```
mgolovko, line 2: Doesn't work in `Modules`. Only in main target. `CoreEnvironment.targetType` works this.
Also needs to setup  `CoreEnvironment.targetType` in all modules that using `SharedContentManager`. It will be better to setup in all targets)

joe-pusya: Remake with `CoreEnvironment.targetType` and user-defined property in build settings

mgolovko (approved):

PR #3194, opened 2025-04-23
IOS-4510 ToastBarData | more SwiftUI way API

mgolovko:
1. ToastBarData is optional now. Remove `showSnackBar` field and `empy` value
2. Changed constructor to shorten syntax
Old:
`ToastBarData(text: "Some text", messageType: .someType)`
New
`ToastBarData("Some text", type: .someType)`

### DIFF
--- a/Anytype/Sources/PresentationLayer/TextEditor/Views/UndoRedo/UndoRedoViewModel.swift
+++ b/Anytype/Sources/PresentationLayer/TextEditor/Views/UndoRedo/UndoRedoViewModel.swift
@@ -10,7 +10,7 @@ final class UndoRedoViewModel: ObservableObject {
     
     private let objectId: String
 
-    @Published var toastData: ToastBarData = .empty
+    @Published var toastData: ToastBarData?
     
     init(
         objectId: String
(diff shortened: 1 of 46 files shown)


Anytype/Sources/PresentationLayer/TextEditor/Views/UndoRedo/UndoRedoViewModel.swift
```diff
@@ -23,7 +23,7 @@ final class UndoRedoViewModel: ObservableObject {
         do {
             try await objectActionsService.undo(objectId: objectId)
         } catch let error as ObjectActionsServiceError {
-            toastData = ToastBarData(text: error.localizedDescription, showSnackBar: true, messageType: .none)
+            toastData = ToastBarData(error.localizedDescription, type: .none)
```
ignatovv, line 26: Seems like it's worth renaming .none to .default now in the current naming.

ignatovv (approved):

Anytype/Sources/PresentationLayer/TextEditor/Views/UndoRedo/UndoRedoViewModel.swift
mgolovko: Maybe rename to `neutral`? `success` type by default.

PR #3199, opened 2025-04-24
Ios-4502 vault redesign

ignatovv:
image


Modules/Assets/Sources/Assets/Resources/Assets.xcassets/DesignSystem/x32/add_filled.imageset/Contents.json
```diff
@@ -0,0 +1,12 @@
+{
+  "images" : [
+    {
+      "filename" : "add_filled.pdf",
+      "idiom" : "universal"
```
mgolovko, line 5: please mark "Render as template"

mgolovko (approved):

PR #3232, opened 2025-04-30
IOS-4516 Space hub | Add simple messaging preview

ignatovv:
Screenshot 2025-04-30 at 12 47 01 PM

### DIFF
--- a/Anytype/Sources/PresentationLayer/Modules/Chat/Services/Models/ChatMessagePreview.swift
+++ b/Anytype/Sources/PresentationLayer/Modules/Chat/Services/Models/ChatMessagePreview.swift
@@ -3,4 +3,37 @@ struct ChatMessagePreview: Hashable {
     let chatId: String
     let unreadCounter: Int
     let mentionCounter: Int
+    
+    let lastMessage: String
+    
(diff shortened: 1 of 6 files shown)


Anytype/Sources/PresentationLayer/Modules/Chat/Services/Models/ChatMessagePreview.swift
```diff
@@ -3,4 +3,37 @@ struct ChatMessagePreview: Hashable {
     let chatId: String
     let unreadCounter: Int
     let mentionCounter: Int
+    
+    let lastMessage: String
+    
+    func updated(unreadCounter: Int, mentionCounter: Int) -> ChatMessagePreview {
```
mgolovko, line 9: probably overhead. You can use `var` properties in ChatMessagePreview

mgolovko (approved):

Anytype/Sources/PresentationLayer/Modules/Chat/Services/Models/ChatMessagePreview.swift
ignatovv: Will fix in next iteration

PR #3241, opened 2025-05-06
IOS-4516 Message Previews | Updated data model

ignatovv:


Anytype/Sources/PresentationLayer/Modules/Chat/Services/ChatMessagesPreviewsStorage.swift
```diff
@@ -106,12 +107,19 @@ actor ChatMessagesPreviewsStorage: ChatMessagesPreviewsStorageProtocol {
...
         guard let subscriptionId, data.subIds.contains(subscriptionId) else { return false }
         
         var preview = previewsBySpace[spaceId] ?? ChatMessagePreview(spaceId: spaceId, chatId: contextId)
         preview.lastMessage = data.message.message.text
-        preview.attachments = data.message.attachments
+
+        let attachmentsIds = data.message.attachments.map(\.target)
+        
+        do {
+            preview.attachments = try await searchService.searchObjects(spaceId: spaceId, objectIds: attachmentsIds)
```
joe-pusya, line 119: maybe we should check for attachmentsIds emptiness to avoid redundant search

joe-pusya (approved):

PR #3246, opened 2025-05-07
iOS-4568 Updated onboarding flow

joe-pusya:
https://github.com/user-attachments/assets/8735a6f6-b38e-438d-88bd-a7f33f1e53e1


Anytype.xcodeproj/project.pbxproj
```diff
@@ -1312,7 +1312,7 @@
 					"$(inherited)",
 					"@executable_path/Frameworks",
 				);
-				MARKETING_VERSION = 0.37.6;
+				MARKETING_VERSION = 0.37.7;
```
ignatovv, line 1315: Seems like it is not suppose to be here

ignatovv (approved):

Anytype.xcodeproj/project.pbxproj
joe-pusya: I increase the version to make a new release build - should be there)

PR #3249, opened 2025-05-07
iOS-4586 Do not show push permissions alert for local-only users

joe-pusya:


ignatovv (approved):

Anytype/Sources/ServiceLayer/PushNotifications/PushNotificationsAlertHandler/PushNotificationsAlertHandler.swift
```diff
@@ -29,6 +30,10 @@ final class PushNotificationsAlertHandler: PushNotificationsAlertHandlerProtocol
             return false
         }
         
+        guard serverConfigurationStorage.currentConfiguration() != .localOnly else {
```
ignatovv, line 33: nit: we have isLocalOnly property in config

ignatovv (approved):

PR #3265, opened 2025-05-12
IOS-4046 Message previews | Handle dependencies from preview

ignatovv:


Anytype/Sources/PresentationLayer/Modules/Chat/Services/Models/ChatMessagePreview.swift
```diff
@@ -2,7 +2,7 @@ import Services
 import Foundation
 
 struct LastMessagePreview: Hashable {
-    let creator: Participant?
+    let creator: ObjectDetails?
```
mgolovko, line 5: You can create Participant from ObjectDetails for use localName

mgolovko (approved):

PR #3292, opened 2025-05-15
IOS-4514 Support new property creation from empty state

ignatovv:
https://github.com/user-attachments/assets/2514d148-0e21-4e2f-bd1b-d86ae81ccf57


Anytype/Sources/PresentationLayer/Common/SwiftUI/Search/ObjectSearch/ObjectSearchView.swift
```diff
@@ -1,6 +1,7 @@
 import Foundation
 import SwiftUI
 
+@available(*, deprecated)
```
mgolovko, line 4: Please add comment, which module we should to use

mgolovko (approved):

PR #3337, opened 2025-05-21
IOS-4654 Core | add mask for sensitive data in logs

mgolovko:


Modules/Services/Sources/Services/Auth/AuthMiddleService.swift
```diff
@@ -38,7 +38,7 @@ final class AuthMiddleService: AuthMiddleServiceProtocol {
     public func createWallet(rootPath: String) async throws -> String {
         let result = try await ClientCommands.walletCreate(.with {
             $0.rootPath = rootPath
-        }).invoke()
+        }).invoke(responseMask: { $0.mnemonic = "***" })
```
ignatovv, line 41: I suggest considering adding this mask inside the invocation itself, for example via code generation. 
This way, if in the future we call these methods from somewhere else, we will bulletproof against someone forgetting to use the mask in a new place.

mgolovko: Need to think about it. It's more complicated.

PR #3473, opened 2025-06-16
IOS-4755 Chat | fix navigation

mgolovko:


Anytype/Sources/Models/Extensions/ObjectDetailsExtensions.swift
```diff
@@ -66,7 +66,7 @@ extension BundledRelationsValueProvider {
         case .bookmark:
             return .bookmark
         case .chat, .chatDerived:
-            return FeatureFlags.chatLayoutInsideSpace ? .chat : .page
+            return .chat
```
mgolovko, line 69: Have a filter for searching

PR #3483, opened 2025-06-17
IOS-4752 Chat | add object open

mgolovko:


Anytype/Sources/Models/Documents/Events/DocumentViewModelSetter.swift
```diff
@@ -62,27 +62,7 @@ final class DocumentViewModelSetter: DocumentViewModelSetterProtocol {
     // MARK: - Private
     
     private func buildBlocksTree(information: [BlockInformation], rootId: String, container: any InfoContainerProtocol, layout: DetailsLayout?) {
-        
         information.forEach { container.add($0) }
-        let roots = information.filter { $0.id == rootId }
```
mgolovko, line 67: We have objects without BlockInformation. Chat object and some other ([Sentry](https://sentry.anytype.io/organizations/anytype/issues/15378)).

PR #3507, opened 2025-06-25
IOS-4783 Common | Delete CachedAsyncImage & Add custom implementation

mgolovko:
CachedAsyncImage works very bad and make a lot of glitches with big images


Anytype/Sources/PresentationLayer/Modules/GalleryInstallationPreview/Subviews/GalleryInstallationPreviewManifestView.swift
```diff
@@ -15,7 +14,7 @@ struct GalleryInstallationPreviewManifestView: View {
                 Spacer.fixedHeight(24)
                 TabView {
                     ForEach(manifest.screenshots, id: \.self) { url in
-                        CachedAsyncImage(
+                        AsyncImage(
```
mgolovko, line 17: This is ok. Cache here is not needed

PR #3544, opened 2025-07-02
IOS-4830 Fix disappearing focus

ignatovv:


Anytype/Sources/PresentationLayer/TextEditor/EditorPage/EditorPageController.swift
```diff
@@ -316,10 +316,6 @@ extension EditorPageController: EditorPageViewInput {
         for item in notExistingItems {
             guard let itemId = item.id else { continue }
             guard let oldItem = snapshot.itemIdentifiers.first(where: { $0.id == itemId }) else {
-                anytypeAssertionFailure(
-                    "Not found old item in snapshot",
-                    info: ["item": String(describing: item)]
-                )
```
ignatovv, line 322: Legit case for block link, removed assert

PR #3562, opened 2025-07-07
iOS-4837 Rename fields

mgolovko:


joe-pusya (approved):
Looks like wrong into branch - I see fixes with loading indicators here

PR #3656, opened 2025-07-23
IOS-4815 Use wrapper view for publishing to handle data load

ignatovv:
https://github.com/user-attachments/assets/804b9c39-579f-4808-a84a-08bf7a93279c

https://github.com/user-attachments/assets/55032513-72b8-4497-8fe1-884e4165a8a6

### DIFF
--- a/Modules/Services/Sources/Services/Publishing/PublishingService.swift
+++ b/Modules/Services/Sources/Services/Publishing/PublishingService.swift
@@ -7,7 +7,7 @@ public protocol PublishingServiceProtocol: Sendable {
     func remove(spaceId: String, objectId: String) async throws
     func list(spaceId: String) async throws -> [PublishState]
     func resolveUri(uri: String) async throws -> PublishState
-    func getStatus(spaceId: String, objectId: String) async throws -> PublishState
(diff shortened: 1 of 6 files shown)


Modules/Services/Sources/Services/Publishing/PublishingService.swift
```diff
@@ -48,12 +48,17 @@ public final class PublishingService: PublishingServiceProtocol {
...
+    public func getStatus(spaceId: String, objectId: String) async throws -> PublishState? {
+        do {
+            let response = try await ClientCommands.publishingGetStatus(.with {
+                $0.spaceID = spaceId
+                $0.objectID = objectId
+            }).invoke(ignoreLogErrors: .badInput, .noSuchObject, .noSuchSpace, .null)
+            
+            return PublishState(from: response.publish)
+        } catch let error as Anytype_Rpc.Publishing.GetStatus.Response.Error {
+            if error.code == .null { return nil }
```
mgolovko, line 60: This logic have in `Invocation`

ignatovv: I hope it is temporary crutch until Tolya will create a dedicated error code for non existent published pages.

PR #3662, opened 2025-07-24
IOS-4908 Implement publishing screen in object settings - Part 2

ignatovv:
Prepare proper data for web previews


Anytype/Sources/PresentationLayer/Modules/PublishToWeb/PublishToWebViewModel.swift
```diff
@@ -35,6 +39,15 @@ final class PublishToWebViewModel: ObservableObject {
             return
         }
         
+        let document = documentsProvider.document(objectId: objectId, spaceId: spaceId)
+        guard let objectDetails = document.details else {
```
mgolovko, line 43: To ensure that the data is exists, need to call `document.open()`. There is a waiting for an existing opening task

PR #3733, opened 2025-08-06
iOS-4980 Core Subscriptions | add preparation for working with types and relationships without open space

mgolovko:
Load types and relationships for cases when we work with them without opening the space:
- Sharing Extension
- Read default object type without open a space
- In next PR will be fixed QuickActions

https://github.com/user-attachments/assets/4a53c597-ee25-433d-8a9c-a55dedcc4d57

### DIFF
--- a/Anytype/Sources/PresentationLayer/Modules/SharingExtensionShareTo/SharingExtensionShareToViewModel.swift
+++ b/Anytype/Sources/PresentationLayer/Modules/SharingExtensionShareTo/SharingExtensionShareToViewModel.swift
@@ -8,8 +8,11 @@ final class SharingExtensionShareToViewModel: ObservableObject {
     private var workspacesStorage: any WorkspacesStorageProtocol
     @Injected(\.searchService)
     private var searchService: any SearchServiceProtocol
+    @Injected(\.activeSpaceManager)
+    private var activeSpaceManager: any ActiveSpaceManagerProtocol
     
     private let data: SharingExtensionShareToData
+    private var dataPrepared = false
     
     @Published var title: String = ""
     @Published var searchText: String = ""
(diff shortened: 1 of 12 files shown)


joe-pusya (approved):

Anytype/Sources/ServiceLayer/ActiveSpaceManager/ActiveSpaceManager.swift
```diff
@@ -98,6 +99,11 @@ actor ActiveSpaceManager: ActiveSpaceManagerProtocol, Sendable {
         workspaceInfoStorage.value = nil
     }
     
+    func prepareSpaceForPreview(spaceId: String) async {
+        await objectTypeProvider.prepareData(spaceId: spaceId)
+        await propertyDetailsStorage.prepareData(spaceId: spaceId)
+    }
```
ignatovv, line 105: I'm not sure this method is really obvious in what it's doing because like when I read prepare space for preview I expect I also can retrieve objects not only types and properties

mgolovko: You can make anything with space. This method prepare global state for work with objects

ignatovv (approved):

Anytype/Sources/PresentationLayer/Modules/SharingExtensionShareTo/SharingExtensionShareToViewModel.swift
```diff
@@ -21,6 +24,12 @@ final class SharingExtensionShareToViewModel: ObservableObject {
     }
     
     func search() async {
+        
+        if !dataPrepared {
+            await activeSpaceManager.prepareSpaceForPreview(spaceId: data.spaceId)
+            dataPrepared = true
+        }
```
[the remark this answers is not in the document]
mgolovko: Fixed. Made a method is optimized for call each times.

PR #3777, opened 2025-08-12
IOS-4905 Add camera permission check before opening camera or document scanner

ignatovv:
## Summary
- Added camera permission check before opening camera or document scanner in the editor
- Shows permission alert if camera access is not granted
- Follows existing pattern from QRCodeScannerViewModifier

## Implementation Details
- Added showCameraPermissionAlert method to EditorPageModuleOutput protocol
- Implemented routing through EditorRouter to EditorPageCoordinatorViewModel
- Camera permission alert is shown when user tries to use camera/scanner without permission

### DIFF
--- a/Anytype/Sources/PresentationLayer/Flows/EditorPageFlow/EditorPageCoordinatorView.swift
+++ b/Anytype/Sources/PresentationLayer/Flows/EditorPageFlow/EditorPageCoordinatorView.swift
@@ -66,5 +66,6 @@ struct EditorPageCoordinatorView: View {
             }
             .snackbar(toastBarData: $model.toastBarData)
             .openUrl(url: $model.openUrlData)
+            .cameraPermissionAlert(isPresented: $model.showCameraPermissionAlertData)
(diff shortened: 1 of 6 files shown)


Anytype/Sources/PresentationLayer/Flows/EditorPageFlow/EditorPageCoordinatorView.swift
```diff
@@ -66,5 +66,6 @@ struct EditorPageCoordinatorView: View {
             }
             .snackbar(toastBarData: $model.toastBarData)
             .openUrl(url: $model.openUrlData)
+            .cameraPermissionAlert(isPresented: $model.showCameraPermissionAlertData)
```
mgolovko, line 69: For any Swiftui view
```
view.cameraAccessFullScreenCover(item: $model.cameraData) {
    SimpleCameraView(data: $0)
}
```
Needs only joining camera view and uikit editor via callbacks

PR #3827, opened 2025-08-25
iOS-5066 New auth | Add new join flow (based on prev navigation)

joe-pusya:
Do not update email collection screen yet

https://github.com/user-attachments/assets/c001c602-4d30-4fd5-926b-b48034fcac9f


Anytype/Sources/PresentationLayer/AuthNew/Join/JoinView/JoinViewModel.swift
```diff
@@ -0,0 +1,87 @@
+import SwiftUI
+
+@MainActor
+final class JoinViewModel: ObservableObject, JoinBaseOutput {
```
joe-pusya, line 4: based on previous JoinFlow navigation (almost the same)

PR #3832, opened 2025-08-26
iOS-5067 Launch View | support light mode

mgolovko:
https://github.com/user-attachments/assets/cf32e6a8-f13b-4135-b97f-870e7d549137


Anytype/Resources/SystemAssets.xcassets/LaunchColors/circle.colorset/Contents.json
```diff
@@ -0,0 +1,38 @@
+{
+  "colors" : [
```
mgolovko, line 2: Colors should be in main target for works in LaunchScreen.storyboard

PR #3838, opened 2025-08-26
IOS-5079 Fixed empty search state

ignatovv:
https://github.com/user-attachments/assets/43e8bd44-47c9-4ab1-bc3b-d1d3c65c51c4


e1sordo:
Hi! 👋

First off, sorry to bother you with this small UI issue in the middle of another PR!

I noticed a tiny thing when using the app: when you tap the input field to search for a space and the keyboard opens, the list of spaces doesn't get a bottom offset equal to the keyboard's height. This makes it impossible to scroll to the very bottom items while the keyboard is active.

Really sorry to sneak this into an unrelated PR, but would you mind taking a quick look at this when you have a moment? I'd be super grateful if you could consider a fix for it.

Thanks a million for all your hard work! 🙏

PR #3847, opened 2025-08-27
IOS-5058 Auth | Add circle with animation

mgolovko:
https://github.com/user-attachments/assets/2a344f9e-9638-4a1f-919c-239a9f990ecf


joe-pusya (approved):
🔥🔥 🔥

PR #3851, opened 2025-08-27
Ios 5020 Remove create object button from widgets

ignatovv:
image


Modules/Assets/Sources/Assets/Resources/Assets.xcassets/DesignSystem/x32/Favorite/Favorite.imageset/Contents.json
```diff
@@ -1,7 +1,7 @@
 {
   "images" : [
     {
-      "filename" : "Pin.svg",
+      "filename" : "State=Inactive.svg",
```
joe-pusya, line 4: Pin name looks better 🤝

PR #3996, opened 2025-09-26
iOS-5205 Type widgets | implement default initial order for type widgets & update type sort in app

mgolovko:


Anytype/Sources/PresentationLayer/Modules/HomeWidgets/Container/HomeWidgetsViewModel.swift
```diff
@@ -102,7 +103,10 @@ final class HomeWidgetsViewModel: ObservableObject {
     }
     
     func typesDropFinish(from: DropDataElement<ObjectTypeWidgetInfo>, to: DropDataElement<ObjectTypeWidgetInfo>) {
-        Task {
+        typesDropTask?.cancel()
```
mgolovko, line 106: task will be deleted if middleware fixes orderId events

PR #4686, opened 2026-03-18
IOS-5899 Korean text does not render italicised on iOS

kmordan:
## Summary
- CJK fallback fonts on iOS (Apple SD Gothic Neo, etc.) lack italic variants, causing Korean/Chinese/Japanese text to appear normal even when italic markup is applied
- Replace font-trait italic with synthetic italic via `.obliqueness` attribute for consistent rendering across all scripts
- Maintain backwards compatibility: removing italic also resets font trait for pre-existing content saved with the old approach

## Notes
- Chat module uses a separate italic rendering path (`font.italic()` in `MessageTextBuilder` / `ChatTextViewCoordinator`) which is not addressed in this PR — follow-up needed

### DIFF
--- a/Anytype/Sources/Design system/TextComponents/Markup/MarkStyleModifier.swift
+++ b/Anytype/Sources/Design system/TextComponents/Markup/MarkStyleModifier.swift
@@ -80,10 +80,14 @@ final class MarkStyleModifier {
             
             
         case .italic:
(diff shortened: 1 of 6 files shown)


ignatovv (approved):
Seems spooky, I hope you know what you are doing

PR #4753, opened 2026-03-30
IOS-5976 Use _final_score for global search and mentions search

requilence:
## Summary
- Added `_final_score` as the primary sort key — a blended backend score combining `ln(1 + BM25) + recency + name_boost`, where recency is exponential decay (30-day half-life) on `lastOpenedDate`/`lastModifiedDate`
- `lastOpenedDate` kept as fallback (won't be needed in practice since recency is already factored in)
- Regenerated middleware with new `BundledPropertyKey.finalScore` case


requilence:
No need for integration on the iOS side. If the client passes fulltext field but doesn't pass _score in order, it is automatically prepended.

PR #4942, opened 2026-05-08
IOS-5994 Detect thermal/memory pressure and forward MW debug profiles to Sentry

requilence:
## Summary
- Wire iOS side of the new middleware debug-profile/Sentry-report integration ([heart #3115](https://github.com/anyproto/anytype-heart/pull/3115), desktop reference [anytype-ts #2174](https://github.com/anyproto/anytype-ts/pull/2174)).
- `SystemMonitor` watches `ProcessInfo.thermalState` + `DispatchSource` memory pressure and asks middleware to run a profile via `runProfiler(reason, reasonDesc)`. Cooldowns: 5 min for timed (thermal) profiles, 30 s for instant (memory pressure).
- New `DebugProfileEventHandler` listens for `Anytype_Event.Debug.ProfileCreated` (emitted by middleware after our trigger or autonomously for `MEMORY_GROWTH`/`LONG_RPC`/`DB_CORRUPTION`) and forwards the artifact to Sentry as a `MW_<reason>` event with the file as attachment, `jsonInfo` parsed into `info` context, and tag/fingerprint matching the desktop client.
- `MetricsService.setInitialParameters` now passes `workdir` so middleware knows where to write profiles; `DebugService` exposes the new `runProfiler / exportReport / cleanupReport` API surface.


Anytype/Sources/ApplicationLayer/AppConfigurators/Configurators/GlobalServicesConfiguration.swift
```diff
@@ -16,7 +16,10 @@ final class GlobalServicesConfiguration: AppConfiguratorProtocol {
     private var appVersionUpdateService: any AppVersionUpdateServiceProtocol
     @Injected(\.debugService)
     private var debugService: any DebugServiceProtocol
-    
+
+    private let systemMonitor = SystemMonitor()
```
kmordan: SystemMonitor and DebugProfileEventHandler should be injected through DI to match project convention, even if it's barely visible that we would ever have different implementations of them.

Anytype/Sources/CoreLayer/Sentry/DebugProfileEventHandler.swift
```diff
@@ -0,0 +1,74 @@
+import Foundation
+import Sentry
+import Services
+import ProtobufMessages
+import Combine
+import Logger
+
+private let log = EventLogger(category: "DebugProfileEventHandler")
```
kmordan: should be placed inside DebugProfileEventHandler

Anytype/Sources/CoreLayer/Sentry/DebugProfileEventHandler.swift
```diff
@@ -0,0 +1,74 @@
...
+                    path: profile.path,
+                    filename: url.lastPathComponent,
+                    contentType: contentType(for: url)
+                ))
+            }
+        }
+    }
+}
+
+private func contentType(for url: URL) -> String {
```
kmordan: We can make a private extension for the URL and put this function inside that extension.

Anytype/Sources/CoreLayer/Sentry/DebugProfileEventHandler.swift
```diff
@@ -0,0 +1,74 @@
...
+    private func forwardToSentry(_ profile: Anytype_Event.Debug.ProfileCreated) {
+        let reason = profile.reason.isEmpty ? "Unknown" : profile.reason
+        log.debug("DebugProfileCreated: reason=\(reason), path=\(profile.path), full=\(profile.full)")
+
+        let event = Event(level: .info)
+        event.message = SentryMessage(formatted: "MW_\(reason)")
+        event.tags = ["report": "mw_profile", "reason": reason]
+        event.fingerprint = ["mw-profile", reason]
+
+        SentrySDK.capture(event: event) { scope in
```
kmordan: It needs to be checked if it is possible to make SentrySDK a dependency within this class.

Anytype/Sources/CoreLayer/Sentry/DebugProfileEventHandler.swift
```diff
@@ -0,0 +1,74 @@
+import Foundation
+import Sentry
+import Services
+import ProtobufMessages
+import Combine
```
kmordan, line 4: do we need this import?

Anytype/Sources/CoreLayer/Sentry/DebugProfileEventHandler.swift
```diff
@@ -0,0 +1,74 @@
...
+
+@MainActor
+final class DebugProfileEventHandler {
+
+    // WORKAROUND: Force linker to retain Anytype_Event.Debug metadata.
+    // Mirrors MembershipStatusStorage workaround — nested type metadata can be
+    // stripped in Release builds when never referenced directly.
+    private static let _forceParentTypeRetention: Anytype_Event.Debug.Type = Anytype_Event.Debug.self
+
+    private var subscription: AnyCancellable?
```
kmordan, line 25: It seems that we need to add "stopSubscriptionAndClean", which will be called on logout and will cancel the subscription.

Anytype/Sources/CoreLayer/Sentry/SystemMonitor.swift
```diff
@@ -0,0 +1,87 @@
+import Foundation
+import Services
+import Factory
+import Logger
+
+private let log = EventLogger(category: "SystemMonitor")
```
kmordan: move inside class

Anytype/Sources/CoreLayer/Sentry/SystemMonitor.swift
```diff
@@ -0,0 +1,87 @@
...
+                } else if event.contains(.warning) {
+                    self.triggerProfiler(durationInSeconds: 0, reason: .memoryPressureWarn, desc: "iOS memory pressure: warning")
+                }
+            }
+        }
+        source.resume()
+        memoryPressureSource = source
+    }
+
+    private func triggerProfiler(durationInSeconds: Int32, reason: DebugProfilerReason, desc: String) {
```
kmordan: not sure we need to use Int32, instead of just Int

Anytype/Sources/CoreLayer/Sentry/SystemMonitor.swift
```diff
@@ -0,0 +1,87 @@
+import Foundation
+import Services
+import Factory
+import Logger
+
+private let log = EventLogger(category: "SystemMonitor")
+
+@MainActor
+final class SystemMonitor {
```
kmordan: The class name is too abstract. We should make it more concrete or perhaps even decompose it into two independent services.

Anytype/Sources/CoreLayer/Sentry/SystemMonitor.swift
```diff
@@ -0,0 +1,87 @@
+import Foundation
+import Services
+import Factory
+import Logger
+
+private let log = EventLogger(category: "SystemMonitor")
+
+@MainActor
```
kmordan: Do we really need to make it a `MainActor`, not just an `actor`? It doesn't have functions that should be called on the main thread, so perhaps making it just an `actor` and calling it in the background would be a good idea.

Anytype/Sources/CoreLayer/Sentry/DebugProfileEventHandler.swift
```diff
@@ -0,0 +1,74 @@
+import Foundation
+import Sentry
+import Services
+import ProtobufMessages
+import Combine
+import Logger
+
+private let log = EventLogger(category: "DebugProfileEventHandler")
+
+@MainActor
```
kmordan: same question is it need to be mainactor.
Do we really need to make it a `MainActor`, not just an `actor`? It doesn't have functions that should be called on the main thread, so perhaps making it just an `actor` and calling it in the background would be a good idea.

Anytype/Sources/CoreLayer/Sentry/SystemMonitor.swift
```diff
@@ -0,0 +1,87 @@
...
+final class SystemMonitor {
+
+    @Injected(\.debugService)
+    private var debugService: any DebugServiceProtocol
+
+    private var memoryPressureSource: DispatchSourceMemoryPressure?
```
kmordan: It seems that we need to add "stopSubscriptionAndClean", which will be called on logout and will cancel the subscription.

Anytype/Sources/CoreLayer/Sentry/SystemMonitor.swift
```diff
@@ -0,0 +1,87 @@
...
+            name: ProcessInfo.thermalStateDidChangeNotification,
+            object: nil
+        )
+    }
+
+    @objc private func thermalStateDidChange() {
+        let state = ProcessInfo.processInfo.thermalState
+        switch state {
+        case .serious:
+            triggerProfiler(durationInSeconds: 30, reason: .thermalSerious, desc: "iOS thermal state: serious")
```
kmordan: Ideally, it would be nice to pass `desc` as an enum in order not to duplicate strings.
something like: `.thermalState(.serious)` , `.memoryPressure(.critical)`

Anytype/Sources/PresentationLayer/Debug/Flows/PublicDebug/PublicDebugMenuViewModel.swift
```diff
@@ -92,6 +92,15 @@ final class PublicDebugMenuViewModel: ObservableObject {
     func debugStat() async throws {
         shareUrlFile = try await debugService.debugStat()
     }
+
+    func generateReport(full: Bool) async throws {
+        try await localAuthWithContinuation(reason: "Generate debug report") { [weak self] in
+            guard let self else { return }
+            let dir = FileManager.default.createTempDirectory().path
+            let result = try await debugService.exportReport(dir: dir, full: full)
```
kmordan, line 100: Is `debugService` not captured by a strong reference? I don't think it should be, but it's worth checking.

Modules/Services/Sources/Services/Debug/DebugService.swift
```diff
@@ -10,6 +10,20 @@ public enum DebugRunProfilerState: Codable {
...
+    case memoryPressureWarn
+    case memoryPressureCritical
+    case thermalSerious
+    case thermalCritical
+}
+
+public struct DebugReportResult: Sendable {
+    public let path: String
+    public let summary: String
+    public let lastModifiedTs: Int64
```
kmordan: not sure we need use Int64, instead of just Int

Modules/Services/Sources/Services/Debug/DebugService.swift
```diff
@@ -22,6 +36,10 @@ public protocol DebugServiceProtocol: AnyObject, Sendable {
 
     @MainActor var debugRunProfilerData: AnyPublisher<DebugRunProfilerState, Never> { get }
     func startDebugRunProfiler()
+
+    func runProfiler(durationInSeconds: Int32, reason: DebugProfilerReason, reasonDesc: String) async
```
kmordan: Int?

kmordan (changes requested):
i will fix it

Anytype/Sources/CoreLayer/Sentry/SystemMonitor.swift
```diff
@@ -0,0 +1,83 @@
+import Foundation
+import Services
+import Factory
+import Logger
+
+private let log = EventLogger(category: "SystemMonitor")
+
+@MainActor
+final class SystemMonitor {
+
```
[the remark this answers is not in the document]
kmordan: Addressed. `SystemMonitor` removed in c3eb8003d7; replacements are `actor`s, which don't need `nonisolated init`.

Anytype/Sources/ApplicationLayer/AppConfigurators/Configurators/GlobalServicesConfiguration.swift
kmordan: Done in c3eb8003d7 (thermal + memory triggers) and 2e59ec40b6 (`DebugProfileEventHandler`). All three are now resolved via `Container.shared` factories in `ServicesDI` and lifecycle-managed by `LoginStateService.startSubscriptions`/`stopSubscriptions`.

Anytype/Sources/CoreLayer/Sentry/DebugProfileEventHandler.swift
kmordan: Investigated. `SentrySDK` is a class with an all-static API (no instance to inject). The codebase has no wrapper protocol around it — `SentryNonFatalLogger` and `SentryConfigurator` also call it directly. Wrapping just for this one call site felt premature, so I left the direct call. Happy to introduce `SentryClientProtocol` (`capture(event:configure:)`) + singleton impl if you'd rather we do it.

Anytype/Sources/CoreLayer/Sentry/SystemMonitor.swift
```diff
@@ -0,0 +1,83 @@
...
+        }
+    }
+
+    private func subscribeToMemoryPressure() {
+        let source = DispatchSource.makeMemoryPressureSource(eventMask: [.warning, .critical], queue: .main)
+        source.setEventHandler { [weak self] in
+            guard let self else { return }
+            let event = source.data
+            if event.contains(.critical) {
+                self.triggerProfiler(durationInSeconds: 0, reason: .memoryPressureCritical, desc: "iOS memory pressure: critical")
```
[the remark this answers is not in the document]
kmordan: Addressed. `SystemMonitor` was split into `MemoryPressureProfilerTrigger` (actor) in c3eb8003d7. The `setEventHandler` closure now captures `[weak self]` only and hops into the actor via `Task { await self?.handleEventFire() }`. Further hardened in 928d08b205 to satisfy Swift 6 strict concurrency.

Anytype/Sources/CoreLayer/Sentry/DebugProfileEventHandler.swift
kmordan: Done in 2e59ec40b6 — moved to `private extension URL { var debugProfileContentType: String }`.

Anytype/Sources/CoreLayer/Sentry/DebugProfileEventHandler.swift
kmordan: Done in 2e59ec40b6 — `private static let log` lives inside the actor now.

kmordan (approved):

PR #5019, opened 2026-06-11
IOS-6409 Preferred space aware cold start

requilence:
## Summary
- Pass `preferredSpaceId` to `Rpc.Account.Select` on cold start so anytype-heart loads only the target space (plus tech space) eagerly: a pending deeplink / push-tap space id wins, otherwise empty (today's behavior). Push-tap cold starts capture the deeplink synchronously from `UIScene.ConnectionOptions.notificationResponse` at scene connection, guaranteeing availability before AccountSelect (with dedup against the later `didReceive` delivery of the same response).
- Call the new `Rpc.Account.PreloadRemainingSpaces` once per process, 2s after the cold start destination screen is shown (Space Hub or the deeplink target); errors swallowed — heart has a 10s timer fallback.
- Everything is gated behind the `preferredSpaceOnColdStart` feature flag (off in release, on in debug); flag off keeps the AccountSelect request byte-identical to today.

### DIFF
--- a/Anytype/Sources/ApplicationLayer/AppCoordinator/ApplicationCoordinatorViewModel.swift
+++ b/Anytype/Sources/ApplicationLayer/AppCoordinator/ApplicationCoordinatorViewModel.swift
@@ -2,6 +2,7 @@ import SwiftUI
 @preconcurrency import Combine
 import AnytypeCore
 import Services
+import DeepLinks
 
 @MainActor
 @Observable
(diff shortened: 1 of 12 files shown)


kmordan:
@requilence question on the cold-start preload.

In `SpaceHubCoordinatorViewModel.setup()` we call `schedulePreload()` in the no-action branch (Space Hub is the cold-start destination). In that path `AccountSelect` got an empty `preferredSpaceID`.

What does an empty `preferredSpaceID` do on `AccountSelect`?
- If it keeps the old behavior and loads all spaces eagerly, then `preloadRemainingSpaces()` in that branch is a no-op and we can drop the call.
- If it also defers the spaces, the call is needed and we keep it.

Want to confirm before we rely on it.

kmordan (approved):

PR #4705, opened 2026-03-22
IOS-0000 feat: Add Picture option to Gallery view Image preview settings

ricokareem:
---

- [x] I understand that contributing to this repository will require me to agree with the [CLA](https://github.com/anyproto/.github/blob/main/docs/CLA.md)

---

### Description

Adds missing "Picture" option to mobile Image preview settings (just like desktop). When selected, displays object picture in Gallery view.

Changes:
- Add Picture enum case to SetViewSettingsImagePreviewCover
- Implement pictureCoverType() for explicit picture display
- Add automatic bookmark picture fallback for backward compatibility
- Add Loc.picture localization key

Fixes issue where bookmark collections configured with "Picture" setting on desktop displayed only generic icons on mobile.

### What type of PR is this? (check all applicable)

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] 🔥 Performance Improvements
- [x] ✅ Test
- [x] 🤖 Build
- [x] 🔁 CI

### Related Tickets & Documents

_There currently is no  public-facing ISSUES tab for iOS/Swift,  only for Android/Kotlin_

#### Mobile & Mobile Gallery Layout Menu (BEFORE)

mobile-before
mobile-menu-before

#### Mobile & Mobile Gallery Layout Menu (AFTER)

mobile-after
mobile-menu-after

#### Desktop & Desktop Gallery Layout Menu
* For comparison*

desktop
desktop-menu

### Added tests?

- [x] 👍 yes
- [ ] 🙅 no, because they aren't needed
- [ ] 🙋 no, because I need help

### Added to documentation?

- [ ] 📜 README.md
- [ ] 📓 [tech-docs](https://github.com/anyproto/tech-docs)
- [x] 🙅 no documentation needed

PR #5023, opened 2026-06-15
IOS-0000 Picture cover option for Gallery views + remove bookmark auto-fallback

kmordan:
## Summary
- Keep the explicit "Picture" cover option for Gallery views (resolves the object's `picture` image, matching anytype-ts desktop), with the resolver reusing the existing `findCover` helper.
- Remove the bookmark picture auto-fallback that overrode the user's explicit "None"/relation cover choice and diverged from desktop.
- Replace placeholder `#expect(true)` tests with real `coverType` tests (picture path, no-fallback regression, non-gallery) and revert unrelated dependency bumps in `Package.resolved`.


kmordan:
Closing as duplicate of #4705. The review fixes have been pushed directly onto the external contributor's PR (#4705) instead.

PR #5074, opened 2026-07-18
IOS-6592 Refresh live cell after paste 'Plain text' so it survives the identity fork

requilence:
### Symptom
Paste a URL into an **empty** or the trailing **"tap to type"** block: it shows as a URL link by default and the "Paste as" menu appears. Choosing **Bookmark** works. Choosing **Plain text** does **nothing** — the block keeps the link markup.

### Root cause
The eager `setNewText(newTextWithLink)` shown before the menu **forks the empty block** (`BlockReplace`) or **materializes the virtual placeholder** (`BlockCreate`), replacing the block with a new id that is rendered by a **freshly built** `TextBlockViewModel`/`TextBlockActionHandler` (the builder caches handlers by id, so the new id misses the cache). The "Paste as" option callbacks still run on the **original pre-fork handler**. `pasteAsText` writes the plain text to the new id correctly (model + middleware update), but its `resetSubject` only reaches the original handler's now-replaced (dead) cell — the live cell for the new id is a different handler and is never refreshed. Text `.setText` local events are `.unhandled` by design (they update `infoContainer` but never reconfigure the text cell; UI refresh flows only through the per-handler `resetSubject`). So the visible forked cell keeps the link. Bookmark/object work because they change the block **type**, forcing a full cell rebuild.

### Fix
After `pasteAsText`'s write, reconfigure the **live** cell for the current (possibly forked/materialized) block id via `modelsHolder.blocksMapping` + `collectionController.reconfigure` — the same idiom `VirtualTrailingBlockSession.applyFocus` already uses. This covers both divergence cases (identity fork of an empty real block, and virtual-placeholder materialization). The other options are unaffected: `pasteAsLink` is a no-op (link already shown), `createBookmark`/`object`/`embed` change the block type and rebuild on their own.

### Keystroke-overwrite guard
`collectionController.reconfigure` re-runs the cell configuration, and `TextBlockContentView.applyNewConfiguration` sets `textView.textStorage.setAttributedString(...)`, so the reconfigure also **re-syncs the live text view** to the plain value. This closes the secondary hole: without it, the next keystroke would read the stale link text back from the text view and write it over the plain model. The write is a **local in-process gRPC** call (embedded `Lib.xcframework`), so its continuation (the reconfigure) runs before any human keystroke after the menu dismisses; the re-sync then guarantees subsequent input reads the plain value.

### Tests
No unit test was added. The paste/fork flow is deeply UIKit-coupled (real `UITextView`, popover menu, a private menu-option closure capturing the pre-fork handler, and a diffable-data-source reconfigure), there is no test precedent that constructs `TextBlockActionHandler`, and the original identity-fork change (IOS-6572) also shipped without handler unit tests. The UI-refresh behavior is an integration concern that a unit test cannot meaningfully assert without brittle scaffolding.

### Build
`xcodebuild build -project Anytype.xcodeproj -scheme Anytype -destination 'generic/platform=iOS Simulator' -configuration Debug CODE_SIGNING_ALLOWED=NO` — **succeeded** (exit 0; only pre-existing warnings unrelated to the change).


requilence:
Checked the `.createBookmark`/`.bottom` path against the fork mechanics — the claimed scenario is unreachable, so leaving line 497 as plain `setNewText`.

1. **The described paste can't open the menu.** `shouldCreateBookmark` guards `replacementText.trimmed.isValidURL()` (line 451): the "Paste as" menu only appears when the pasted string is a *bare* URL. Pasting `"Check this out: https://…"` fails the guard — no menu, no callback.

2. **`position == .bottom` and the identity fork are mutually exclusive.** `.bottom` requires the text view to hold more than the pasted URL, i.e. the block had pre-existing content. But then the eager `setNewText(newTextWithLink)` cannot fork: `canForkEmptyBlockOnFill` requires `content.text.isEmpty`, and a virtual placeholder the user typed into materialized on the first keystroke — after which the focus carry-over (IOS-6572) moves editing to the live cell, so the paste flow starts on the live handler. With no in-flow identity change, `self` *is* the live handler and the line-497 restore's `resetSubject` reaches the live cell. No stale UI.

3. **In the two divergence cases the restore never executes.** When the eager set does fork (empty real block) or materialize (untouched placeholder), the block was empty, so `newTextWithLink` is exactly the bare linked URL. `setNewText` sends `resetSubject` synchronously *before* `carryOrSync` (line 811), so the captured `textView` holds exactly `trimmedText` when the option callback runs → `position == .replace` → the line 494–498 restore is skipped entirely (and `.replace` destroys the source block anyway). This matches the tested behavior: Bookmark works in both divergence cases.

The only way to line up ".bottom + diverged handler" is pasting a URL within the few-ms window while the first-keystroke fork/materialization RPC is still in flight — a race that predates this PR and isn't reachable by human input. Routing 497 through `setNewTextRefreshingLiveBlock` would be harmless but would imply a reachable divergence that doesn't exist.

PR #5076, opened 2026-07-20
IOS-6579 Restore collapsing object header on Kanban board

requilence:
## Summary
- Restores the whole-page collapsing object header on Board layout (lost by the scroll virtualization restructure): every column reserves the collapse distance with a top spacer, the actively scrolled column drives the shared `headerTravel` by its scroll **deltas**, and the settings row + `SetFullHeader` are overlays slaved to it — dragging anywhere collapses the header first, then scrolls cards, matching Grid/List/Gallery, with the header sliding fully off-screen like the other view types. New `KanbanCollapseCoordinator` keeps columns in sync (finger-only drivership + fling-stop on claim, two-way page lock so the all-group-headers state is always recoverable, deep columns keep their position, over-drag bounces only the dragged column, observers unregister on window detach); column header moved inside the column scroll as a pinned opaque section header. TODO IOS-6595 tracks moving the KVO/introspection plumbing to iOS 18 scroll APIs.
- Fixes card drag-and-drop, broken on the virtualized board: the axis-lock gate never resolved on a stationary press, blocking the drag lift (and its haptic) for the whole touch — it now fails itself after 0.3s of a stationary hold. The whole column is now a drop target (previously releasing over the top spacer, pinned header, card gaps or the empty bottom area performed no drop — card stayed dimmed and the move silently never persisted), a background target no longer teleports a card to the end of its own column, and the drag dim engages only once a real drop target is entered so phantom `onDrag` payload invocations can't leave a dropped card greyed out.
- Adds a "Move to" context menu on cards: stationary long-press → menu listing the other columns by name; moving during the press still lifts into the normal drag. Reuses the exact drag-drop move pipeline (optimistic move, persistence, revert, analytics); hidden without move permissions or on single-column boards.

Design doc + field-test log: `docs/plans/20260720-ios-6579-kanban-collapsing-header-handoff.md`

### DIFF
--- a/Anytype/Sources/PresentationLayer/Common/SwiftUI/BasicComponents/ScrollAxisLock.swift
+++ b/Anytype/Sources/PresentationLayer/Common/SwiftUI/BasicComponents/ScrollAxisLock.swift
@@ -66,6 +66,7 @@ private final class ScrollAxisLockUIView: UIView {
 private final class CrossAxisDragGate: UIGestureRecognizer {
     private let scrollAxis: Axis
     private var startLocation = CGPoint.zero
+    private var stationaryHoldTimeout: DispatchWorkItem?
 
     init(scrollAxis: Axis) {
         self.scrollAxis = scrollAxis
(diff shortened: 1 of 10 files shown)


requilence:
Tested on devices

kmordan (approved):

PR #5083, opened 2026-08-04
IOS-6597 Fix autocorrect dropping the first letter in new text blocks

requilence:
## Summary

- The first character typed into a middleware-empty text block swapped its block id (BlockReplace fork, or BlockCreate for the trailing placeholder). Diffable identity derives from the block id, so the row was delete+inserted, the focused `UITextView` died and the keyboard input session restarted — autocorrect then computed suggestions from the word minus its first letter. The row now keeps a stable identity and its view model rebinds in place, so the swap diffs as an empty change and the same text view keeps first responder.
- Undo of a swap rebinds the row back to the restored block. All of it behind `stableRowIdentityOnFork` (default on); with the flag off the previous pipeline runs unchanged.
- `.at` focus positions past the end of the text clamp to the end instead of collapsing to 0, and paste re-applies its caret once the pasted characters arrive — previously the caret jumped to the start of the block after a paste.

### DIFF
--- a/AnyTypeTests/PresentationLayer/EditorPage/BlockRowIdentityMapTests.swift
+++ b/AnyTypeTests/PresentationLayer/EditorPage/BlockRowIdentityMapTests.swift
@@ -1,0 +1,72 @@
+import Testing
+@testable import Anytype
+
+@MainActor
+struct BlockRowIdentityMapTests {
+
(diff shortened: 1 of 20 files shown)


kmordan (approved):

requilence:
closed in favor of #5084

